Issue #5, opened 2016-04-12
Use of UUID instead of Advertising ID


nseidm1:
generateSiteUuid() retrieves the device UUID which changes upon factory reset, and is different for each user of the device. The Advertising ID is the standard and expected way to identify a particular user.

http://developer.android.com/google/play-services/id.html

emmett9001:
Thanks @nseidm1. I'll prioritize this issue for the next few weeks.

mypplication:
So? When can you update your SDK with another solution ?
Thx

cwisecarver:
Fixed by #15. Will be included in the next release.

PR #15, opened 2017-10-05
Use Google Play Services to get the advertising id

hbbtstar:
This should address https://github.com/Parsely/parsely-android/issues/5.

Confirmed working in the ParselyExample app (which has been updated due to that testing).

I should point out there is a possibility, if a track call fires extremely quickly, that the adKey is still null for the first pageview: but it's a consequence of getting the IdInfo in an asynctask, which the Android docs recommend. We could block on getting the ad info, but that seems like a bad solution to me.

I think the rare possibility of one pageview having a different uuid is worth it, considering how hard it was to replicate in code (the adkey get took just a couple hundred milliseconds), but if we really want to ensure that doesn't happen the only other way I can think of other than blocking on that call (which we shouldn't do) it is to loop through the queue and replace the deviceInfo if adKey is set before we flush it.

### DIFF
--- a/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/parselyandroid/ParselyTracker.java
@@ -343,6 +343,7 @@ public class ParselyTracker {
         this.flushInterval = flushInterval;
         this.storageKey = "parsely-events.ser";
         //this.rootUrl = "http://10.0.2.2:5001/";  // emulator localhost
+        this.rootUrl = "http://srv.pixel.parsely.com/";
         this.urlref = urlref;
         this.queueSizeLimit = 50;
         this.storageSizeLimit = 100;
--- a/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/parselyandroid/ParselyTracker.java
@@ -492,14 +492,16 @@ public class ParselyTracker {
             String advertId = null;
             try {
                 idInfo = AdvertisingIdClient.getAdvertisingIdInfo(mContext);
-            } catch (GooglePlayServicesRepairableException |IOException|GooglePlayServicesNotAvailableException e) {
+            }
+            catch (GooglePlayServicesRepairableException | IOException | GooglePlayServicesNotAvailableException e) {
                 PLog("No Google play services or error! falling back to device uuid");
                 // fall back to device uuid on google play errors
                 advertId = getSiteUuid();
             }
             try {
                 advertId = idInfo.getId();
-            } catch (NullPointerException e) {
+            }
+            catch (NullPointerException e) {
                 advertId = getSiteUuid();
             }
             return advertId;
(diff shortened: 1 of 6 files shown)


parsely/parselyandroid/ParselyTracker.java
```diff
@@ -331,10 +337,12 @@ protected ParselyTracker(String apikey, int flushInterval, String urlref, Contex
 
         this.apikey = apikey;
         this.uuidkey = "parsely-uuid";
+        this.adKey = null;
+        // get the adkey straight away on instantiation
+        new GetAdKey(c).execute();
         this.flushInterval = flushInterval;
         this.storageKey = "parsely-events.ser";
         //this.rootUrl = "http://10.0.2.2:5001/";  // emulator localhost
-        this.rootUrl = "http://srv.pixel.parsely.com/";
```
emmett9001: Why is this removed? Looks like it's still important.

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -469,4 +477,38 @@ protected Void doInBackground(Void... params) {
...
+            mContext = context;
+        }
+
+        @Override
+        protected String doInBackground(Void... params) {
+            AdvertisingIdClient.Info idInfo = null;
+            String advertId = null;
+            try {
+                idInfo = AdvertisingIdClient.getAdvertisingIdInfo(mContext);
+            } catch (GooglePlayServicesRepairableException |IOException|GooglePlayServicesNotAvailableException e) {
```
emmett9001: Strange spacing around operators here

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -469,4 +477,38 @@ protected Void doInBackground(Void... params) {
...
+                PLog("No Google play services or error! falling back to device uuid");
+                // fall back to device uuid on google play errors
+                advertId = getSiteUuid();
```
emmett9001: This works without `this` because this class is nested within `ParselyTracker`, right?

parsely/parselyandroid/ParselyTracker.java
hbbtstar: it definitely is, not sure how I axed it and then missed it! adding back.

parsely/parselyandroid/ParselyTracker.java
hbbtstar: good catch, will make it neater

parsely/parselyandroid/ParselyTracker.java
hbbtstar: yup, inner classes can access outer class variables like this (unless I had declared a variable in the inner class with the same name)

hbbtstar:
in testing some more I was unable to get the adKey to not set even firing a PV as soon as the app loads (I suspect Google Play Services caches this) so I think we're good to go without doing all the complicated checking in the batch header. Updated with other comments!

PR #16, opened 2017-11-14
Catch NullPointerException occurring in stopFlushTimer

cwisecarver:
My reading of the android traceback in [this ticket](https://github.com/Parsely/web/issues/7005#issuecomment-344055353) is that there's a `NullPointerException` bubbling out of `ParselyTracker.stopTimerFlush`. I couldn't recreate it. The customer didn't really give me any information other than it's happening most often on Samsung devices and in Android 7. I looked at all the events in datadog from August 10th when the customer reported the problem starting, nothing stood out. I investigated the java.util.Timer class. It looks to me like calling `purge` after calling `cancel` and then setting `this.timer` to `null` is duplicative. Both setting timer to `null` and calling `purge` dereference the queue, timer, and it's tasks so that GC will collect them.  I did find alternative opinions saying that you should call `purge` after `cancel`. So this seemed like the optimal solution. Purge where you can, if it fails catch the exception, log, set `this.timer` to `null`, and move on.

### DIFF
--- a/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/parselyandroid/ParselyTracker.java
@@ -282,7 +282,11 @@ public class ParselyTracker {
     public void stopFlushTimer(){
         if(this.timer != null){
             this.timer.cancel();
-            this.timer.purge();
+            try {
+                this.timer.purge();
+            } catch (NullPointerException ex) {
+                PLog("Exception caught during Timer.purge(): %s", ex.toString());
+            }
         }
         this.timer = null;
     }


emmett9001:
This looks like what I was imagining, and I'm glad you found some documentation to back up this approach. I'll let you do the merge.

cwisecarver:
@emmett9001 What's the release process for this work? What are the next steps to get this into the hands of the Toronto Star so they can see if it resolves their issue? Thanks for the review!

emmett9001:
We'll want to cut another [bugfix version](https://github.com/Parsely/parsely-android/releases), then notify the customer that they can update their apps. This project is distributed through github at the moment. In the future it will be worth investigating whether there are some Android package indices that we can distribute on - I'm not very familiar with the ecosystem myself.

PR #19, opened 2019-02-19
Code auto-formatting.

kbourgoin:
Ran the Android Studio tools to clean up the code. Nothing fancy, but wanted this as a separate PR before I move on to miscellaneous finishing touches.

Depends on #18. Merge that first and this will look more reasonable.

PR #20, opened 2019-02-19
Some finishing touches.

kbourgoin:
A whole mess of comments in here and a quick bugfix. I'm debating more extensive changes, but this will do for now. I still need to update readmes and make sure the docs here will generate correctly, but that can be a separate PR if y'all are on a review-spree tomorrow.

Depends on #18 and #19 to make the diff sensible.


kbourgoin:
@emmett9001 base changed on PR

parsely/parselyandroid/ParselyMetadata.java
```diff
@@ -5,51 +5,66 @@
...
  */
 public class ParselyMetadata {
     public ArrayList<String> authors, tags;
-    public String canonical_url, section, thumbUrl, title;
+    public String canonicalUrl, section, thumbUrl, title;
     public Calendar pubDate;
 
+    /* \brief Create a new ParselyMetadata object.
+     *
+     * @param authors         List of authors for the article.
```
emmett9001: We use the terminology "post or page" in our [metadata documentation](https://www.parse.ly/help/integration/jsonld/). I think "article" isn't the most futureproof.

parsely/parselyandroid/ParselyMetadata.java
```diff
@@ -5,51 +5,66 @@
...
-        this.canonical_url = canonical_url;
+        this.canonicalUrl = canonicalUrl;
         this.section = section;
         this.tags = tags;
         this.thumbUrl = thumbUrl;
         this.title = title;
         this.pubDate = pubDate;
     }
 
+    /* \brief Turn the ParselyVideoMetadata into a Map
```
emmett9001: I'm not a fan of using the class name in the documentation of one of its members.
> Turn this object into a Map

Also, the class in this file is `ParselyMetadata`, not `ParselyVideoMetadata`.

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -152,55 +152,92 @@ public static ParselyTracker sharedInstance(String apikey, int flushInterval, St
...
+     *                     content). Do not use for URLs that Parse.ly would normally crawl.
      */
     public void trackURL(String url, ParselyMetadata urlMetadata) {
         this.enqueueEvent(this.buildEvent(url, "pageview", urlMetadata));
     }
 
+    /*! \brief Start engaged time tracking for the given URL.
+     *
+     * This starts a timer which will send events to Parse.ly on a regular basis
+     * to capture engaged time for this URL. This URL should match the value passed
```
emmett9001: Imprecise - which call to `trackURL` does a call to this function need to match? Suggestion:
> The URL passed to this function should be a URL for which `trackURL` has been called.

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -210,26 +247,38 @@ public void startEngagement(String url) {
...
-    /*! \brief Add an event Map to the queue.
+    /*! \brief Stop engaged time tracking.
      *
-     *  Place a data structure representing the event into the in-memory queue for later use
-     *
-     *  **Note**: Events placed into this queue will be discarded if the size of the persistent queue
-     *  store exceeds `storageSizeLimit`.
+     * Stops the engaged time tracker, sending any accumulated engaged time to Parse.ly.
+     * NOTE: This *must* be called during various Android lifecycle events like onPause or
+     * onStop. Otherwise, engaged time tracking will keep running and Parse.ly values
```
emmett9001: Note: on iOS there's a [way](https://github.com/Parsely/AnalyticsSDK-iOS/blob/0761abfd50f8889fcce536b581ffe41a07c3c31d/ParselyTracker/ParselyTracker.swift#L185) to set up this behavior automatically. We treated that as a blocker for release, though it's arguable whether it actually is.

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -210,26 +247,38 @@ public void startEngagement(String url) {
...
     }
 
+    /*! \brief Start video tracking.
+     *
+     * Starts engaged time tracking for a video. Will send a video start event unless the
+     * same video had previously been paused. Video metadata must be provided, specifically
+     * the video ID and video duration.
+     *
+     * The URL value is used for videos embedded in other posts. This is because a single
+     * video can be embedded in multiple articles, but that is more common on full webpages.
```
emmett9001: "full webpages" -> "web pages"

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -210,26 +247,38 @@ public void startEngagement(String url) {
...
+     * Most apps do not embed videos in articles, but instead play them full screen. In that
```
emmett9001: I don't think I've seen a mobile browser or app that plays videos in a non-fullscreen view. However, I still feel that having a video fullscreened doesn't preclude it being logically "embedded" in some other content, in this case an app view. I'd rephrase this docstring to indicate that the url can be a real url representing the embedding content or a "real-looking" url representing a dummy page that will make it through the backend's allowed-domains checks.

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -210,26 +247,38 @@ public void startEngagement(String url) {
...
+     * case, the value of URL should be the same as the video ID.
```
emmett9001: My takeaway from a conversation from this morning was that in the cases described here, `url` should be a "real-looking" dummy URL and should not necessarily match the video ID.

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -258,18 +307,27 @@ public void trackPlay(String url, @NonNull ParselyVideoMetadata videoMetadata) {
         this.videoEngagementManager.start();
     }
 
+    /*! \brief Pause video tracking.
+     *
+     * Stops engagement tracking an ongoing video. If `trackPlay` is immediately called again for
```
emmett9001: missing a "for" - "for an ongoing video"

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -258,18 +307,27 @@ public void trackPlay(String url, @NonNull ParselyVideoMetadata videoMetadata) {
...
-            PLog("No ongoing video to stop.");
             return;
         }
-        PLog("trackPause called");
         this.videoEngagementManager.stop();
     }
 
+    /*! \brief Reset tracking on a video.
+     *
+     * Stops engaged time tracking and resets state for video tracking. This means that if
```
emmett9001: Unnecessary distinction here between "engaged time tracking" and "video tracking"
> Stops video tracking and resets internal state.

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -278,11 +336,13 @@ public void resetVideo() {
 
     /*! \brief Create an event Map
      *
-     *  @param url The canonical URL identifying the pageview/heartbeat
-     *  @param action Action kind to use (e.g. pageview, heartbeat)
+     *  @param url      The canonical URL identifying the pageview/heartbeat
+     *  @param action   Action kind to use (e.g. pageview, heartbeat)
```
emmett9001: "Action kind" is inaccurate - I'd understand the "action kinds" to be `pageview` and `heartbeat`, where the events of kind "pageview" are `pageview` and `videostart` events. I think you mean "action name" or simply "action"

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -310,9 +370,18 @@ public void resetVideo() {
...
+     *
+     *  **Note**: Events placed into this queue will be discarded if the size of the persistent queue
+     *  store exceeds `storageSizeLimit`.
+     *
+     *  @param event The event Map to enqueue.
+     */
     private void enqueueEvent(Map<String, Object> event) {
         // Push it onto the queue
-        PLog("%s", event);
+        // PLog("%s", event);
```
emmett9001: Delete instead of commenting

parsely/parselyandroid/ParselyTracker.java
```diff
@@ -360,13 +428,20 @@ private void sendBatchRequest(ArrayList<Map<String, Object>> events) {
...
+     * @return Whether the network is accessible and Parsely is reachable.
+     */
     private boolean isReachable() {
         ConnectivityManager cm = (ConnectivityManager) this.context.getSystemService(
                 Context.CONNECTIVITY_SERVICE);
         NetworkInfo netInfo = cm.getActiveNetworkInfo();
         return netInfo != null && netInfo.isConnectedOrConnecting();
     }
 
+    /*! \brief Save the event queue to permanent storage.
```
emmett9001: "permanent" -> "persistent"

parsely/parselyandroid/ParselyMetadata.java
kbourgoin: :+1:

parsely/parselyandroid/ParselyMetadata.java
kbourgoin: Woopsie there. I'll change both things about it.

parsely/parselyandroid/ParselyTracker.java
kbourgoin: I looked briefly at Android activity lifecycle and it looks like only the main activity has access to these events. I'll add it to my list to poke around more and see if we can make it automatic.

PR #21, opened 2019-04-22
implement proportional backoff for engagement tracking

emmett9001:
This pull request implements Parsely's standard heartbeat backoff mechanism in the EngagementManager.


emmett9001:
This is blocked by the equivalent change in the closed-source Parsely JS SDK.

PR #31, opened 2021-07-13
Remove assertion and replace will nullcheck

cwisecarver:
In an attempt to fix [this issue](https://github.com/Parsely/web/issues/11886) this removes the assertion and replaces it will a null check, if that fails it returns an empty queue.

### DIFF
--- a/ParselyExample/build.gradle
+++ b/ParselyExample/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         google()
     }
     dependencies {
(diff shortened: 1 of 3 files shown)


kbourgoin (approved):
This looks like exactly the fix we need. Thank you for putting this together.

emmett9001 (approved):

cwisecarver:
Happy to merge this and comment on the issue if that's the move from here.

emmett9001:
Works for me

Issue #35, opened 2022-06-13
Introduce `example` and `lib` modules


wzieba:
We need to divide the project into 2 modules: SDK (library) and example app in order to offer only library code from the repository.

PR #36, opened 2022-06-13
chore: introduce example and lib modules

wzieba:
\+ remove unused files.

Closes: #35 

### Description

This PR creates two modules: `parsely` and `example` so the structure of the project is:

```
Root project 'parsely-android'
+--- Project ':example'
\--- Project ':parsely'
```

`example` module contains a sample app and `parsely` - the SDK, which will be offered on MavenCentral.

---

I've also removed a few lines of unused code in this PR but not all of it - I'll handle it in a separate PR.


ParaskP7 (approved):
👋 @wzieba !

Really nice work on restructuring the project so that it builds from the root, right after someones clones it, and also removing the duplicate `com.parsely.parselyandroid` files, effectively making it a library instead.

I cloned the project and made it work with the old structure, I then pointed to this PR's branch and did the same with the new structure. Everything worked as expected. Thanks also for cleaning up the unnecessary testing related structure, some build configuration and a few of the IDE files that should be committed to the codebase anyway.

Everything looks so much better now and works as expected. It is a 👍 from my side!

-----

- My testing mainly involved running `clean`, `assembleDebug` and `adb install` to install the app. Everything worked as expected and with no issues. 💯 
- In addition to that, I also tryied to run `assembleRelease` but as you would expect as well I got a failure due to the `targetSdkVersion 28` that is no longer support. Which, to be honest is why you are doing all this work anyway. Thus, I am sure you are aware of it. 🥇 
- I would have added other minor comments related to for example `ProGuard`, extra `Gradle` configuration, etc, but at this point I too think it is just better to progress anyway and do an overall clean-up/re-organization at the very end and in case we need to, that is after this project is updated to the point we feel comfortable with it. 👍

wzieba:
Thanks a lot for the review @ParaskP7 🙏 !

> I would have added other minor comments related to for example ProGuard, extra Gradle configuration

Some of the extra Gradle configurations should be probably resolved in the next PRs but I agree that we can do overall clean-up at the very end 👍 .

Issue #37, opened 2022-06-13
Set up a CI for PR validation


wzieba:
To make development and releases easier and more robust, we should introduce a CI.

PR #38, opened 2022-06-13
ci: configure Github Actions basic build

wzieba:
Closes: #37


.github/workflows/readme.yml
```diff
@@ -0,0 +1,20 @@
+name: Assemble project
+
+on:
+  push:
```
wzieba, line 4: The SDK is not under active development and it's open source. The build time is also very short. I don't think we can reach any limits with the `push` trigger and it allows us to work freely with feature branches.

ParaskP7 (approved):
LGTM! 💯

PR #40, opened 2022-06-13
chore: bump AGP and Gradle Wrapper

wzieba:
\+ use Gradle Plugin DSL 

Closes: #39


settings.gradle
```diff
@@ -1 +1,17 @@
+pluginManagement {
```
ParaskP7, line 1: Praise (❤️): Adding `Plugin DSL` is the right move, kudos! 🥇

settings.gradle
```diff
@@ -1 +1,17 @@
...
+    repositories {
```
ParaskP7, line 2: Suggestion (💡): Consider moving as much of the `Plugin DSL` related configuration to the `settings.gradle` file. For example, you could end up with the below:
```
pluginManagement {
    gradle.ext.agpVersion = '7.2.1'

    plugins {
        id 'com.android.application' version gradle.ext.agpVersion
        id 'com.android.library' version gradle.ext.agpVersion
    }

    repositories {
        gradlePluginPortal()
        google()
        mavenCentral()
    }
}
```

Also, with the above you might consider removing the root level `build.gradle` build file as it is no longer necessary. PS: The `clean` task is not necessary and can be usually removed.

settings.gradle
```diff
@@ -1 +1,17 @@
...
+
+dependencyResolutionManagement {
+    repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
+    repositories {
+        google()
+        mavenCentral()
+    }
+}
+
 include ':parsely', ':example'
```
ParaskP7, line 17: Suggestion (💡): You might also consider splitting the `include` into separate lines so that when and if you do an update to any of the modules the change for a specific module will apply to a specific line thus not affecting any other module (just to be on the safe side). Also, with this pattern you don't need to worry about commas (`,`) separating the modules:
```
include ':parsely'
include ':example'
```

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Mon Feb 11 11:31:10 EST 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
```
ParaskP7, line 3: Praise (❤️): Latest of Gradle and AGP, love it! 🥇

ParaskP7 (commented):
👋 @wzieba !

I have reviewed and tested this PR, good job! 🥇 

I have left a couple of suggestions (💡) for you to consider, I am not approving this PR just get to give you some time to think about and potentially apply them. Let me know what you think and we can quickly progress with this PR, nothing is blocking. 👍

settings.gradle
wzieba: Thanks! I've committed this change and removed the root project `build.gradle`. 

I'm not sure about `clean` task - now it doesn't remove the `build` directory of root project when running `clean`. But I think it's a minor thing.

settings.gradle
wzieba: Thanks! I prefer having a single `include` but I've moved declarations into separate lines.

Issue #41, opened 2022-06-15
Set up MavenCentral publishing


wzieba:
We should offer the possibility to use SDK via repository. The most common and reliable is Maven Central.

PR #42, opened 2022-06-15
feat: repository publish

wzieba:
Closes: #41 

### Description

This PR adds configuration for publishing library to Nexus Repository (from which it can be promoted to Maven Central).

It also adds a CI workflow triggered on a new release event.

image

With this configuration, the library has been successfully released to the MavenCentral repository: https://search.maven.org/artifact/com.parsely/parsely/3.0.2/aar

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,20 @@
+# Project-wide Gradle settings.
+
+# IDE (e.g. Android Studio) users:
+# Gradle settings configured through the IDE *will override*
+# any settings specified in this file.
+
+# For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+# Default value: -Xmx10248m -XX:MaxPermSize=256m
+# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+
+# When configured, Gradle will run in incubating parallel mode.
+# This option should only be used with decoupled projects. More details, visit
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true
+
+android.disableAutomaticComponentCreation=true
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -6,7 +6,7 @@ android {
     compileSdkVersion 28
 
     defaultConfig {
-        minSdkVersion 23
+        minSdkVersion 21
         targetSdkVersion 28
         versionCode 1
         versionName "1.0"
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -8,8 +8,6 @@ android {
     defaultConfig {
         minSdkVersion 21
         targetSdkVersion 28
-        versionCode 1
-        versionName "1.0"
     }
     buildTypes {
         release {
@@ -25,12 +23,6 @@ android {
 }
 
 dependencies {
-    implementation 'com.android.support:appcompat-v7:28.0.0'
-    implementation 'com.android.support:customtabs:28.0.0'
-    implementation 'com.android.support:support-v4:28.0.0'
-    implementation 'com.android.support:support-media-compat:28.0.0'
-    implementation 'com.android.support:design:28.0.0'
     implementation 'org.codehaus.jackson:jackson-mapper-lgpl:1.9.13'
-    implementation 'org.codehaus.jackson:jackson-core-lgpl:1.9.13'
     implementation 'com.google.android.gms:play-services-ads:17.1.3'
 }
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -20,9 +20,24 @@ android {
         exclude 'META-INF/LICENSE'
         exclude 'META-INF/NOTICE'
     }
+
+    publishing {
+        singleVariant('release') {
+            withSourcesJar()
+            withJavadocJar()
+        }
+    }
+}
+
+ext {
+    PUBLISH_GROUP_ID = 'com.parsely'
+    PUBLISH_VERSION = '3.0.2'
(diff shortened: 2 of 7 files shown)


gradle.properties
```diff
@@ -0,0 +1,20 @@
...
+# The setting is particularly useful for tweaking memory settings.
+# Default value: -Xmx10248m -XX:MaxPermSize=256m
+# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+
+# When configured, Gradle will run in incubating parallel mode.
+# This option should only be used with decoupled projects. More details, visit
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true
+
+android.disableAutomaticComponentCreation=true
```
wzieba, line 20: This will be a default behavior in AGP 8.0 so I already enforced it for this project. Surprisingly, it made e.g. sharing sources easier. More info: https://developer.android.com/studio/build/build-variants

parsely/build.gradle
```diff
@@ -6,10 +6,8 @@ android {
     compileSdkVersion 28
 
     defaultConfig {
-        minSdkVersion 23
+        minSdkVersion 21
```
wzieba, line 9: There's no point to enforce 23 min SDK version.

publication.gradle
```diff
@@ -0,0 +1,69 @@
...
+                        connection = 'scm:git:github.com/Parsely/parsely-android.git'
+                        developerConnection = 'scm:git:ssh://github.com/Parsely/parsely-android.git'
+                        url = 'https://github.com/Parsely/parsely-android/tree/main'
+                    }
+
+                    developers {
+                        developer {
+                            id = 'emmettbutler'
+                            name = 'Emmett Butler'
+                            email = 'emmett.butler@automattic.com'
```
wzieba, line 53: @emmettbutler MavenCentral enforces adding a developer associated with the project. Is that okay for you?

emmettbutler: Yes, thanks for asking

gradle.properties
ParaskP7: Thanks for adding a comment on that @wzieba !

PS: I am adding [this link](https://developer.android.com/studio/releases/gradle-plugin#disable-component-creation) as well as I wasn't sure when to look on the already share [link above](https://developer.android.com/studio/build/build-variants).

parsely/build.gradle
ParaskP7: 👋 @wzieba !

1) How about decreasing that on the `example` project as well since it is also point to `23` at this point? 🤔 
2) How about extracting these `compileSdkVersion`, `minSdkVersion` and `targetSdkVersion` as `ext` property on a root level `build.gradle` file and share them so they could be updated in one place vs. in multiple places and thus potentially missed?

PS: The root level `build.gradle` file can look as simple as:

```
ext {
    minSdkVersion = 21
    compileSdkVersion = 28
    targetSdkVersion = 28
}

```

parsely/build.gradle
```diff
@@ -22,15 +20,23 @@ android {
...
+        singleVariant('release') {
+            withSourcesJar()
+        }
+    }
+}
+
+ext {
+    PUBLISH_GROUP_ID = 'com.parsely'
+    PUBLISH_VERSION = '3.0.2'
+    PUBLISH_ARTIFACT_ID = 'parsely'
```
ParaskP7, line 34: 1) Question (❓): Why did you chose this `build.gradle` file and location to put these publishing related information?
2) Suggestion (💡): If possible, consider moving them to a newly created file, maybe a root level `publication.properties` file, to match that of the `publication.gradle` file, something like the below. You will also note that I reorder them as well to have the version last and as such match its usage (`com.parsely:parsely:3.0.2`):

`publication.properties`
```
publish.groupId=com.parsely
publish.artifactId= parsely
publish.version=3.0.2

```

publication.gradle
```diff
@@ -0,0 +1,69 @@
...
+apply plugin: 'signing'
+
+group = PUBLISH_GROUP_ID
+version = PUBLISH_VERSION
+
+def prepareEnvironment() {
+    ext["ossrhUsername"] = System.getenv('OSSRH_USERNAME')
+    ext["ossrhPassword"] = System.getenv('OSSRH_PASSWORD')
+    ext["signing.keyId"] = System.getenv('SIGNING_KEY_ID')
+    ext["signing.password"] = System.getenv('SIGNING_PASSWORD')
```
ParaskP7, line 11: Minor (🔍): Consider moving this below the `ext["signing.key"]` in order to match the order within the `useInMemoryPgpKeys`. As you can tell, this is purely an OCD related comment because of consistency reasons, to have the `keyId` and `key` properties grouped, and the password last... 😅

ParaskP7 (commented):
👋 @wzieba !

I have reviewed this PR, good job! 🥇

I have left a question (❓), a suggestions (💡) and a minor (🔍) comment for you to consider, I am not approving this PR just get to give you some time to think about and potentially apply them. Let me know what you think and we can quickly progress with this PR, nothing is blocking. 👍

parsely/build.gradle
ParaskP7: That's for the extra clarification @wzieba !

Moving those to the `publication.gradle` seems a good choice too, 👍 from my side on that!

wzieba: I did yet another change: I've moved group and name of the artifact to publication script but I left `VERSION` in `build.gradle` of the module.

The reasons are:
1. Version is a coherent part of the build (doesn't matter if it's published)
2. It'll be easy to find it later as we'll need to update this version to release a new aar 🙂

I think with all of these we came to the best solution!

ParaskP7: @wzieba coolio, deal, I did check the changes you did and it all LGTM! 👍 💯

parsely/build.gradle
ParaskP7: True, I agree with that, having the `example` app module disconnected from the `parsely` library module in that sense. I'll let this decision up to you then, it seems that you have thought about it, so 👍 from my side!

ParaskP7 (approved):

Issue #45, opened 2022-06-16
Update `jackson` dependency


wzieba:
Currently used `org.codehaus.jackson:jackson-mapper-lgpl:1.9.13` dependency has a security issue: 

https://ossindex.sonatype.org/vulnerability/CVE-2017-7525?component-type=maven&component-name=org.codehaus.jackson%2Fjackson-mapper-lgpl&utm_source=ossindex-client&utm_medium=integration&utm_content=1.7.0

PR #46, opened 2022-06-17
fix: bump Jackson to 2.13.3

wzieba:
Closes: #45 

### Description

This PR updates Jackson (JSON parser) library. The update is forced because of a security flaw of the currently used version ([details](https://ossindex.sonatype.org/vulnerability/CVE-2017-7525?component-type=maven&component-name=org.codehaus.jackson%2Fjackson-mapper-lgpl&utm_source=ossindex-client&utm_medium=integration&utm_content=1.7.0)).

### How I tested the change

I've taken parsed data after pressing all buttons from top to bottom in the example app, before and after the version bump. See `ParselyTracker#JsonEncode` method.

The diff result is the following:
https://www.diffchecker.com/vtzFPlw2

If I interpret the result correctly the changes are related to:
- timestamp ✅
- order of the parsed objects. But `java.util.HashMap` is already a map that does not respect the order of the values so the dependency bump did not change this behavior. ✅

Taking both into account I think that the bump is safe.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -12,11 +12,6 @@ android {
         versionCode 1
         versionName "1.0"
     }
-    packagingOptions {
-        exclude 'META-INF/LGPL2.1'
-        exclude 'META-INF/LICENSE'
-        exclude 'META-INF/NOTICE'
-    }
 }
 
 dependencies {
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -32,7 +32,7 @@ android {
 
 dependencies {
     implementation 'androidx.appcompat:appcompat:1.4.2'
-    implementation 'org.codehaus.jackson:jackson-mapper-lgpl:1.9.13'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.3'
     implementation 'com.google.android.gms:play-services-ads:21.0.0'
 }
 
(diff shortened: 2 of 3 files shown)


ParaskP7 (commented):
👋 @wzieba !

I reviewed and tested this PR as per the description. Everything LGTM! 💯 

Question (❓) I checked on the [diff](https://www.diffchecker.com/vtzFPlw2) as well (btw thanks) and it seems that in addition to the `timestamp` changes, which I also think are expected, I noticed that the before version had the `inc` and `tt` values of the `heartbeat` action pointing to `0`, but the new version had those pointing to `1`. I am not sure if that is something minor or expected, but I thought I should mention anyway, just in case. Wdyt? 😊 

PS: I only reviewed these two extra commits ([this](https://github.com/Parsely/parsely-android/pull/46/commits/5fd0ba84578085ae880467dae755276ec28d0d75) and [this](https://github.com/Parsely/parsely-android/pull/46/commits/4cc0051c02a0b3d8c37535e3b011c836e16f6dcb)). For the rest of the 25 commits it seems that I have already reviewed them. Is that correct? 🤔

wzieba:
> (...) but I thought I should mention anyway, just in case. Wdyt? 😊

Thanks for mentioning it! I've browsed the code and both values are time-based. I think we're good with them being different in the diff.

> PS: I only reviewed these two extra commits ([this](https://github.com/Parsely/parsely-android/pull/46/commits/5fd0ba84578085ae880467dae755276ec28d0d75) and [this](https://github.com/Parsely/parsely-android/pull/46/commits/4cc0051c02a0b3d8c37535e3b011c836e16f6dcb)). For the rest of the 25 commits it seems that I have already reviewed them. Is that correct? 🤔

Sure, thanks. As we talked in DMs, rebase seems to be not interpreted as well as I'd like to on Github 😅 .

PR #48, opened 2022-06-17
docs: update readme to direct to MavenCentral artifacts

wzieba:


wzieba:
hi @emmettbutler 👋  could you take a look at the updated README.md?

And is it okay to remove the `mainpage.dox` file? I haven't found usages of it in this repository but I don't know if it's a part of something else.

emmettbutler (approved):
Looks great. Folks will love how much easier this makes the installation. Removing the .dox file is fine as well.

Issue #57, opened 2023-08-25
NullPointerException when accessing deviceInfo map in ParselyTracker.java


rafayali:
When using 3.0.6 version of SDK, our app is reporting crashes due to `deviceInfo` map not being initialized when accessing its entries when calling `ParselyTracker.trackPageView`. Below is a stacktrace from crashlytics:

Android 13
Parsley SDK 3.0.6

`Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.Object java.util.Map.get(java.lang.Object)' on a null object reference
       at com.parsely.parselyandroid.ParselyTracker.buildEvent(ParselyTracker.java:404)
       at com.parsely.parselyandroid.ParselyTracker.trackPageview(ParselyTracker.java:234)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
       at androidx.compose.ui.platform.AndroidUiDispatcher.performTrampolineDispatch(AndroidUiDispatcher.android.kt:81)
       at androidx.compose.ui.platform.AndroidUiDispatcher.access$performTrampolineDispatch(AndroidUiDispatcher.android.kt:41)
       at androidx.compose.ui.platform.AndroidUiDispatcher$dispatchCallback$1.doFrame(AndroidUiDispatcher.android.kt:68)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1299)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1309)
       at android.view.Choreographer.doCallbacks(Choreographer.java:923)
       at android.view.Choreographer.doFrame(Choreographer.java:847)
       at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:1283)
       at android.os.Handler.handleCallback(Handler.java:942)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8757)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1067)`

[...]

PR #58, opened 2023-09-04
Initialize `deviceInfo` in constructor of `ParselyTracker`

wzieba:
Closes: #57 

## Description

This PR addresses the issue reported in #57 by initializing `deviceInfo` synchronously in the constructor of the `ParselyTracker`, rather than waiting for `GetAdKey` async task to complete.

It brings back part of the behavior from before #56 (which introduced the discussed bug) but makes it more explicitly, that initial creation of `deviceInfo` passes `null` for `adKey` property.

## Reproduction

1. Checkout `trunk`
2. Apply the PATCH

<details>
<summary>PATCH</summary>

```PATCH
Subject: [PATCH] refactor: move plugins configuration to pluginManagement
---
Index: parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
IDEA additional info:
<+>UTF-8
diff --git a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java	(revision 652ccbec41975239f3981289d94a40335f2495b2)
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java	(date 1693839376735)
@@ -738,6 +738,11 @@
         protected String doInBackground(Void... params) {
             AdvertisingIdClient.Info idInfo = null;
             String advertId = null;
+            try {
+                Thread.sleep(10000);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
             try {
                 idInfo = AdvertisingIdClient.getAdvertisingIdInfo(mContext);
             } catch (GooglePlayServicesRepairableException | IOException | GooglePlayServicesNotAvailableException | IllegalArgumentException e) {
``` 

</details>

3. Run the `example` app and within 10 seconds tap on "track url" button.

### DIFF
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -60,7 +60,7 @@ public class ParselyTracker {
     private static int DEFAULT_FLUSH_INTERVAL_SECS = 60;
     private static int DEFAULT_ENGAGEMENT_INTERVAL_MILLIS = 10500;
     protected ArrayList<Map<String, Object>> eventQueue;
-    private String siteId, rootUrl, storageKey, uuidKey, adKey;
+    private String siteId, rootUrl, storageKey, uuidKey;
     private boolean isDebug;
     private SharedPreferences settings;
     private int queueSizeLimit, storageSizeLimit;
@@ -79,7 +79,6 @@ public class ParselyTracker {
 
         this.siteId = siteId;
         this.uuidKey = "parsely-uuid";
-        this.adKey = null;
         // get the adkey straight away on instantiation
         new GetAdKey(c).execute();
         this.storageKey = "parsely-events.ser";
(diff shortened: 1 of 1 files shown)


parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
```diff
@@ -755,9 +755,7 @@ protected String doInBackground(Void... params) {
 
         @Override
         protected void onPostExecute(String advertId) {
-            adKey = advertId;
-            deviceInfo = collectDeviceInfo();
-            deviceInfo.put("parsely_site_uuid", adKey);
```
ParaskP7, line 760: Question (❓): Just to confirm that, why was that deleted? My guess is because it is just being a duplicate of [dInfo.put("parsely_site_uuid", uuid);](https://github.com/Parsely/parsely-android/blob/652ccbec41975239f3981289d94a40335f2495b2/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java#L647), right?

FYI: Because this change was included as part of this 98c1dd6583b71e165ebb1cf2dbeea3bf0e5a85ee refactor commit, which was mainly about removing the `adKey` field, it kind of made it harder for me to understand. Although you did add this on the description, I just want to be sure of it (🤷 + 😊): `It makes codebase easier to understand and eliminates multiple sources for 'parsely_site_uuid'.`

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the `reproduction` instructions, everything works as expected, I was able to crash the app without [this.deviceInfo = collectDeviceInfo(null);](https://github.com/Parsely/parsely-android/pull/58/files#diff-fe50d3e0c151a2e8bc4a2041b1adb1e6f15b770549fa2600c14f0607c9654f76R83), and able to verify that it is now not crashing when this line fix is included, good job, I think we can safely merge this now! 🌟 🌟 🌟

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
wzieba: Yes, this code was duplicating [dInfo.put("parsely_site_uuid", uuid)](https://github.com/Parsely/parsely-android/blob/652ccbec41975239f3981289d94a40335f2495b2/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java#L647) from `collectDeviceInfo`.

Issue #60, opened 2023-09-07
Engagement and video sessions share the same `ts` value of events


wzieba:
### Actual behavior
If a user starts an engagement or video session, the timestamps (`data.ts` parameter) of single heartbeats are the same and equal to the timestamp of the last recorded event.

### Expected behavior
Each heartbeat event should have a distinct `data.ts` value of timestamp at which the event was recorded.

<img width="400" alt="Screenshot 2023-09-07 at 11 47 11" src="https://github.com/Parsely/parsely-android/assets/5845095/7536ae01-b1bd-4d15-a326-26324fcc42a2">

Internal discussion: pd3OIC-1is-p2#comment-313

PR #62, opened 2023-09-07
Copy `data` map when creating a new heartbeat event

wzieba:
Closes: #60 

Instead of referencing to the same `Map<String, Object> data` across all events, when updating timestamps of events in `EngagementManager`, we will now create a shallow copy of `data`. This way will stop introducing unwanted changes of data.ts to existing events.

### Screen from debugger

You can see that after the change, each event has its own `data` HashMap so they can update it independently.

| Before | After |
| --- | --- | 
| Screenshot 2023-09-07 at 14 49 44 | Screenshot 2023-09-07 at 14 50 50 |

### Payload result

You can see that each event has now its own timestamp of record

| Before | After | 
| --- | --- | 
| Screenshot 2023-09-07 at 11 47 11 | Screenshot 2023-09-07 at 11 50 06 |

### DIFF
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -894,11 +894,10 @@ public class ParselyTracker {
             Calendar now = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
             @SuppressWarnings("unchecked")
             Map<String, Object> baseEventData = (Map<String, Object>) event.get("data");
-            if (baseEventData != null) {
-                Map<String, Object> data = new HashMap<>((Map<String, Object>) baseEventData);
-                data.put("ts", now.getTimeInMillis() / 1000);
-                event.put("data", data);
-            }
+            assert baseEventData != null;
+            Map<String, Object> data = new HashMap<>((Map<String, Object>) baseEventData);
+            data.put("ts", now.getTimeInMillis() / 1000);
+            event.put("data", data);
 
             // Adjust inc by execution time in case we're late or early.
             long executionDiff = (System.currentTimeMillis() - scheduledExecutionTime);
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -887,7 +887,7 @@ public class ParselyTracker {
 
         private void doEnqueue(long scheduledExecutionTime) {
             // Create a copy of the base event to enqueue
-            Map<String, Object> event = new HashMap(this.baseEvent);
+            Map<String, Object> event = new HashMap<>(this.baseEvent);
             PLog(String.format("Enqueuing %s event.", event.get("action")));
 
             // Update `ts` for the event since it's happening right now.
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -892,6 +892,7 @@ public class ParselyTracker {
 
             // Update `ts` for the event since it's happening right now.
             Calendar now = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+            @SuppressWarnings("unchecked")
             Map<String, Object> baseEventData = (Map<String, Object>) event.get("data");
             if (baseEventData != null) {
                 Map<String, Object> data = new HashMap<>((Map<String, Object>) baseEventData);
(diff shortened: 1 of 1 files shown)


parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
```diff
@@ -892,8 +892,9 @@ private void doEnqueue(long scheduledExecutionTime) {
 
             // Update `ts` for the event since it's happening right now.
             Calendar now = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-            Map<String, Object> data = (Map<String, Object>) event.get("data");
+            Map<String, Object> data = new HashMap<>((Map<String, Object>) event.get("data"));
```
ParaskP7, line 895: 1. Suggestion (💡): How would you feel about suppressing the `unchecked` warning on `(Map<String, Object>)` to make it explicit that we know what we are doing here?
2. Suggestion (💡): How would you feel about using `requireNonNull(...)` to make it explicit that we expect `event.get("data")` to never be null?
3. Minor (🔍): Not related to your change, but since you correctly used `new HashMap<>` instead of `new HashMap`, consider changing the above such construction from `new HashMap(this.baseEvent)` to `new HashMap<>(this.baseEvent)`.

PS: With the above suggestions and minor improvements I am hoping to end-up with a warning-free `doEnqueue(...)` method. Feel free to ignore or apply any. 💯

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR (see `My Test Instructions` below), everything seems to be working as expected, nice fix! 🌟 

-----

FYI: Because I didn't know how to exactly test this, to verify that everything worked (bug) and works (fix) as expected, I used the below test instructions of mine. Let me know if the that look good to you, or if you tested this in some kind of another way. This will greatly help me understand how to best test future such PRs.

My Test Instructions (using the `example` app):
- Click `START ENGAGEMENT` button.
- Wait 1 second.
- Click `START ENGAGEMENT` button.
- Wait 1 second.
- Wait for the flushing queue to finish (30 seconds).
- Check the `POST Data` and `ts` (JSON).
- Click `STOP ENGAGEMENT` button.
- Wait 1 second.
- Click `STOP ENGAGEMENT` button.
- Wait 1 second.
- Wait for the flushing queue to finish (30 seconds).
- Check the `POST Data` and `ts` (JSON).

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
wzieba: Thank you for the review @ParaskP7 ! 

> 1. Suggestion (💡): How would you feel about suppressing the unchecked warning on (Map<String, Object>) to make it explicit that we know what we are doing here?

Sure 👍 

> 2. Suggestion (💡): How would you feel about using requireNonNull(...) to make it explicit that we expect event.get("data") to never be null?

I've decided to not force non-nullable `event.get("data")` to not crash users applications. It shouldn't ever happen, but I thought it's better safe than sorry in this case (especially for libraries). WDYT?

> 3. Minor (🔍): Not related to your change, but since you correctly used new HashMap<> instead of new HashMap, consider changing the above such construction from new HashMap(this.baseEvent) to new HashMap<>(this.baseEvent).

Sure! Now `doEnqueue` method is warning free. The PR is ready for the second round!

wzieba: > this peace of code was there for years and already tested (by our users)

On the second thought, actually that's a good point 👍 . I've brought back the fail-fast behavior. I used `assert` instead of `requireNonNull` as it's not Kotlin (yet) 😄. https://github.com/Parsely/parsely-android/pull/62/commits/bc62a92deab5452e0085ca419fa2e7b091316cc9

ParaskP7: Awesome, thank @wzieba ! 🚀 

> I've brought back the fail-fast behavior. I used assert instead of requireNonNull as it's not Kotlin (yet) 😄.

`assert` or whichever way you do it is totally fine by me, but just as an FYI `Objects.requireNonNull(event.get("data")` works for Java too. 😅

```java
Map<String, Object> data = new HashMap<>(
        (Map<String, Object>) Objects.requireNonNull(event.get("data"))
);
```

Feel free to merge whenever, no need to change anything! 🚀

wzieba: TIL, thanks! I merged the PR before reading the comment but I'll know this for the future 🙂

randyriback:
Hi there @wzieba -- quick question. Is the Android SDK currently in a place where we can recommend it to customers (who don't have video tracking), or best to wait until this is deployed?

wzieba:
hi @randyriback ! Yes, IMO the current version of Android SDK is in a place where we can recommend it to customers. Saying this, the fix merged in this PR will be deployed today so maybe it's worth to wait just a little bit. I'll ping you here when the new version will be available.

wzieba:
[3.0.8](https://github.com/Parsely/parsely-android/releases/tag/3.0.8) is released @randyriback

Issue #61, opened 2023-09-07
Update time units to milliseconds


wzieba:
Currently, fields `data.ts` and `tt` use seconds as the time unit, while on other platforms - milliseconds are used.

It's not a breaking issue as Parse.ly supports two formats, but it'd be preferable to bring parity between platforms.

Internal discussion: pd3OIC-1is-p2#comment-313

PR #63, opened 2023-09-07
Use milliseconds for `ts` and `tt` parameters

wzieba:
### Description

Closes: #61 

This PR updates time units for `tt` (total time) and `data.ts` (timestamp) parameters. This will match iOS behaviour and was discussed internally here: pd3OIC-1is-p2#comment-313

| Before | After |
| --- | --- | 
| Screenshot 2023-09-07 at 15 50 18 | Screenshot 2023-09-07 at 15 48 21 |

### DIFF
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -404,7 +404,7 @@ public class ParselyTracker {
         data.put("manufacturer", this.deviceInfo.get("manufacturer"));
         data.put("os", this.deviceInfo.get("os"));
         data.put("os_version", this.deviceInfo.get("os_version"));
-        data.put("ts", now.getTimeInMillis() / 1000);
+        data.put("ts", now.getTimeInMillis());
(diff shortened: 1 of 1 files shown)


wzieba:
Thanks for the review @ParaskP7 !

> 1. Question (❓): I understand that on iOS, the data.inc field is using seconds instead of milliseconds. Maybe we should change that as well (to milliseconds), just to have everything using the same time unit (thus forget working/sending with seconds, which might further help with duplicate events), then update that on iOS too (for parity purposes). Wdyt? 🤔

I think that for now, we should keep seconds as iOS followed patterns of JS SDK. Also, `tt` and `inc` are incremented only by seconds, so we don't lose any information here.

> 2. Question (❓): Similar to the above, another usage of seconds is on data.pubDate. I wonder if we can/want to do such change there too? 🤔

Good point! I looked at iOS codebase and they use `pub_date` key for this value, while Android uses `pub_date_tmsp` - I'm not entirely sure what's the difference here (maybe units?) so I'd be extremely careful with updating this value. I'll add an issue to investigate it furter 👍

Issue #28, opened 2021-03-10
Attach a different UUID to every request


kbourgoin:
We've had a couple issues come in where it could be either a bug in a mobile app causing data to be sent more than once, or a bot hitting the data collection endpoint. Bots hitting that endpoint are very unlikely, but the information we send right now makes it hard to be certain. We should attach a diagnostic UUID (or any random number) to each request made so they can be uniquely identified. This would remove at least one source of uncertainty in our investigations.

PR #73, opened 2023-10-04
feat: add internal method for UUID generation

wzieba:
Partially addresses #28 

TBD

### DIFF
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -351,15 +351,15 @@ public class ParselyTracker {
                 videoEngagementManager = null;
             }
         }
-        final @NonNull String uuid = generatePixelId();
+        @NonNull final String uuid = generatePixelId();
 
         // Enqueue the videostart
-        Map<String, Object> videostartEvent = buildEvent(url, urlRef, "videostart", videoMetadata, extraData);
+        @NonNull final Map<String, Object> videostartEvent = buildEvent(url, urlRef, "videostart", videoMetadata, extraData);
         videostartEvent.put(VIDEO_START_ID_KEY, uuid);
         enqueueEvent(videostartEvent);
 
         // Start a new engagement manager for the video.
-        Map<String, Object> hbEvent = buildEvent(url, urlRef, "vheartbeat", videoMetadata, extraData);
+        @NonNull final Map<String, Object> hbEvent = buildEvent(url, urlRef, "vheartbeat", videoMetadata, extraData);
         hbEvent.put(VIDEO_START_ID_KEY, uuid);
         // TODO: Can we remove some metadata fields from this request?
         videoEngagementManager = new EngagementManager(timer, DEFAULT_ENGAGEMENT_INTERVAL_MILLIS, hbEvent);
@@ -412,6 +412,7 @@ public class ParselyTracker {
      * @param extraData A Map of additional information to send with the event.
      * @return A Map object representing the event to be sent to Parse.ly.
      */
+    @NonNull
     private Map<String, Object> buildEvent(
             String url,
             String urlRef,
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -66,6 +66,7 @@ public class ParselyTracker {
 //    private static final String ROOT_URL = "http://10.0.2.2:5001/";
     private static final String ROOT_URL = "https://p1.parsely.com/";
     private static final String UUID_KEY = "parsely-uuid";
+    private static final String VIDEO_START_ID_KEY = "vsid";
 
     protected ArrayList<Map<String, Object>> eventQueue;
     private final String siteId;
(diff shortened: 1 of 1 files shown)


parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
```diff
@@ -349,12 +351,16 @@ public void trackPlay(
                 videoEngagementManager = null;
             }
         }
+        @NonNull final String uuid = generatePixelId();
```
ParaskP7, line 354: Minor (🔍): Although adding `@NonNull` on a local variable doesn't hurt, it is kind of a redundancy as Java is at least smart enough to notice that a local variable maybe produce (or not) a NPE even without such a `@NonNull` or a `@Nullable` annotation. This is especially true since the `generatePixelId()` method is already annotated as `@NonNull`. Same applies to `buildEvent(...)` and the fact that its return type is also annotated as `@NonNull`.

FYI: From my experience these nullability annotation are mostly only used at constructor/method parameter and field level.

PS: Also, using final and a `@NonNull` is a further redundancy as this local variable can't be updated to `null` any longer anyway.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
```diff
@@ -349,12 +351,16 @@ public void trackPlay(
...
 
         // Enqueue the videostart
-        enqueueEvent(buildEvent(url, urlRef, "videostart", videoMetadata, extraData));
+        @NonNull final Map<String, Object> videostartEvent = buildEvent(url, urlRef, "videostart", videoMetadata, extraData);
+        videostartEvent.put(VIDEO_START_ID_KEY, uuid);
```
ParaskP7, line 358: Suggestion (💡): Instead of explicitly using an extra `.put(VIDEO_START_ID_KEY, uuid)` call after the event is constructed and returned, wherever that might be required, and thus letting the caller decide the event's state, which can be prone to misusage, how about moving that logic with the `buildEvent(...)` method itself and avoid any such misusage?

```java
@NonNull
    private Map<String, Object> buildEvent(
            String url,
            String urlRef,
            String action,
            ParselyMetadata metadata,
            Map<String, Object> extraData,
            @Nullable String uuid) {

        ...

        if (action.equals("videostart") || action.equals("vheartbeat")) {
            event.put(VIDEO_START_ID_KEY, uuid);
        }

        return event;
    }
```

FYI: I also think that doing this makes it more visible and explicit and only `videostart` and `vheartbeat` type of events can have an associated `vsid` attached to them.

PS: I understand that this is adding an extra parameter to the method, and that would have been better if a builder pattern would to be used, but I still think it is better than letting callers decide that after the fact, when the even is already created and returned by the `buildEvent(...)` method. 🤔

ParaskP7 (approved):
👋 @wzieba !

-----

I have left a suggestion (💡) and a minor (🔍) comment for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
wzieba: Sure, I can do that 👍 

> I understand that this is adding an extra parameter to the method

That was my concern when deciding on approach here, but keeping all "bulding blocks" in `buildEvent` method is probably indeed better!

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
wzieba: Both `@NonNull` and `@Nullable` annotations have target of `LOCAL_VARIABLE` so I think it's fine using them here. You're right about `uuid` not being nullable is known from `generatePixelId` annotation, but at the same time, I'd have to look at `generatePixelId` signature to know it. Those annotations are not only for IDE, but also for people reading the code. I'd leave this as it is, as it doesn't hurt and helps in code readability IMO.

We can look at this from a different point: Kotlin equivalent of
```java
        @NonNull final String uuid = generatePixelId();
```
is
```
    val uuid: String = generatePixelId();
```

If we removed annotations and `final`, it'd be
```
    var uuid: String? = generatePixelId();
```

ParaskP7: Thanks for the reply @wzieba , as I mentioned this is minor and I am just glad it is being discussed! 🥇 

> ...only that final in this context is redundant!

Ah, actually I was saying that `@NonNull` is redundant, not the `final` keyword, I don't mind that, and to be honest, I kind of like it, doesn't hurt at all.

PS: Another Kotlin equivalent could have been one that you don't get a type reference at all: `val uuid = generatePixelId()`

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
ParaskP7: Thanks for c05858254631ac13f36b9bfa5b0ec384054054a9 @wzieba , but I think you just missed removing this `hbEvent.put(VIDEO_START_ID_KEY, uuid);` bit [here](https://github.com/Parsely/parsely-android/blob/8de8a964124f03ca6c3d41c29224a660f30df1de/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java#L362), right? 🤔

wzieba: Yes, thank you! 🙏

PR #74, opened 2023-10-05
Add UUID to each pageview event

wzieba:
Closes: #28 

### Description

This PR is a twin of #73 for `pageview` and `heartbeat` events. It adds a new, internal field `lastPageviewUuid`. See internal discussion about necessity of it: p1696417656257339-slack-C0533SEJ82H

### Test

1. Install example app, open logcat
2. Tap on `Track Url`
3. Tap on `Start Engagement`
4. Wait 30 seconds until you see `[Parsely] POST Data {"events": (...)` log. Copy the JSON part.
5. Wait another 30 seconds until another `[Parsely] POST Data {"events": (...)` log, copy it as well.
6. Check the copied logs:
- there should be one ` "action": "pageview"` event, rest of them should be `"action": "heartbeat"`
- every event from both logs should have `"pvid"` field with the same value

### New payload

<details>
<summary> Click </summary>
Screenshot 2023-10-05 at 18 34 23

</details>

### DIFF
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
@@ -365,7 +365,6 @@ public class ParselyTracker {
 
         // Start a new engagement manager for the video.
         @NonNull final Map<String, Object> hbEvent = buildEvent(url, urlRef, "vheartbeat", videoMetadata, extraData, uuid);
-        hbEvent.put(VIDEO_START_ID_KEY, uuid);
         // TODO: Can we remove some metadata fields from this request?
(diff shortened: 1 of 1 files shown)


ParaskP7 (approved):
👋 @wzieba !

-----

1. Thinking (🤔): Although I am "kind of" not a fan of having this new `lastPageviewUuid` state added and using it as such, depending that the clients would first `trackPageview(...)` and then `startEngagement(...)`, just because it is in the docs, since I understand that this was discussed and agreed on over Slack, I am okay keeping it as is.
2. Warning (⚠️): Let's deal with [this](https://github.com/Parsely/parsely-android/pull/73#discussion_r1348559473) here and remove it before merging this PR to `master`.

PR #78, opened 2023-10-12
Add unit tests setup

wzieba:
## Description

This PR adds unit tests setup for the SDK. I decided to introduce Robolectric to be able to test `AsyncTask` API.

It also adds support for measuring codecoverage via `kover` plugin, as well as integration with Codecov: https://app.codecov.io/github/Parsely/parsely-android

## How to test
CI should be enough: you could also run `koverHtmlReportRelease` task and see generated HTML report with 0% code coverage.


parsely/build.gradle
```diff
@@ -27,27 +29,38 @@ android {
...
     }
+
+    testOptions {
+        unitTests {
+            includeAndroidResources = true
+        }
+    }
+
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_11.toString()
```
ParaskP7, line 40: Suggestion (💡): I suggests using `VERSION_1_8` for now, just like we are doing across every other lib and client repo (except from [DayOne-Android](https://github.com/search?q=repo%3Abloom%2FDayOne-Android%20JavaVersion.VERSION_11&type=code) 🤷), just to be on the safe side... 🤔

parsely/build.gradle
```diff
@@ -27,27 +29,38 @@ android {
...
-            }
-        }
-        add("implementation", "org.jetbrains.kotlin:kotlin-stdlib-jdk8") {
-            version {
-                require(kotlinVersion)
-            }
-        }
-    }
+    
+    testImplementation 'org.robolectric:robolectric:4.9.2'
```
ParaskP7, line 60: Suggestion (💡): Consider using the latest version of Robolectric ([4.10.3](https://github.com/robolectric/robolectric/releases/tag/robolectric-4.10.3)).

parsely/build.gradle
```diff
@@ -27,27 +29,38 @@ android {
...
+    testImplementation 'androidx.test:core:1.5.0'
+    testImplementation 'org.assertj:assertj-core:3.24.1'
```
ParaskP7, line 62: Suggestion (💡): Consider using the latest version of AssertJ Core ([3.24.2](https://mvnrepository.com/artifact/org.assertj/assertj-core/3.24.2)).

settings.gradle
```diff
@@ -4,6 +4,8 @@ pluginManagement {
     plugins {
         id 'com.android.application' version gradle.ext.agpVersion
         id 'com.android.library' version gradle.ext.agpVersion
+        id 'org.jetbrains.kotlin.android' version '1.9.10'
+        id 'org.jetbrains.kotlinx.kover' version '0.7.4'
```
ParaskP7, line 8: 1. Suggestion (💡): Consider extracting the versions, just like we are doing across every other lib and client repo (ish, but trying). See below:

```gradle
    gradle.ext.agpVersion = '7.2.2'
    gradle.ext.kotlinVersion = '1.9.10'
    gradle.ext.koverVersion = '0.7.4'

    plugins {
        id 'com.android.application' version gradle.ext.agpVersion
        id 'com.android.library' version gradle.ext.agpVersion
        id 'org.jetbrains.kotlin.android' version gradle.ext.kotlinVersion
        id 'org.jetbrains.kotlinx.kover' version gradle.ext.koverVersion
    }
```

2. Suggestion (💡): Since we are on the extracting version world, let me take this opportunity and suggest you to even consider doing the same everywhere and this way reuse any duplicate versions like SDK versions and dependencies like `androidxAppcompatVersion = '1.4.2'` for both the `parsely` and the `example` modules. I such, I suggest creating a new root level `build.gradle` file, just like we are doing across every other lib and client repo (ish, but trying). See below:

Root level `build.gradle` file:
```gradle
ext {
    minSdkVersion = 21
    compileSdkVersion = 33
    targetSdkVersion = 33
}

ext {
    // main
    androidxAppcompatVersion = '1.4.2'
    androidxLifecycleVersion = '2.6.2'
    googlePlayServicesAdsIdentifierVersion = '18.0.1'
    jacksonDatabindVersion = '2.13.3'

    // test
    androidxTestCoreVersion = '1.5.0'
    assertjVersion = '3.24.2'
    junitVersion = '4.13.2'
    robolectricVersion = '4.10.3'
}
```

-----

`parsely` module `build.gradle` file:
```gradle
android {
    compileSdkVersion rootProject.compileSdkVersion

    defaultConfig {
        minSdkVersion rootProject.minSdkVersion
        targetSdkVersion rootProject.targetSdkVersion
    }
    ...
}

dependencies {
    implementation "androidx.appcompat:appcompat:$androidxAppcompatVersion"
    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonDatabindVersion"
    implementation "com.google.android.gms:play-services-ads-identifier:$googlePlayServicesAdsIdentifierVersion"
    implementation "androidx.lifecycle:lifecycle-process:$androidxLifecycleVersion"

    testImplementation "org.robolectric:robolectric:$robolectricVersion"
    testImplementation "androidx.test:core:$androidxTestCoreVersion"
    testImplementation "org.assertj:assertj-core:$assertjVersion"
    testImplementation "junit:junit:$junitVersion"
}
```

-----

`example` module `build.gradle` file:
```gradle
android {
    compileSdkVersion rootProject.compileSdkVersion

    defaultConfig {
        applicationId "com.parsely.example.parselyexample"
        minSdkVersion rootProject.minSdkVersion
        targetSdkVersion rootProject.targetSdkVersion
        versionCode 1
        versionName "1.0"
    }
}

dependencies {
    implementation project(':parsely')
    implementation "androidx.appcompat:appcompat:$androidxAppcompatVersion"
}
```

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, great job, let's start testing! 🌟 🎉 🌟 

-----

> This PR adds unit tests setup for the SDK. I decided to introduce Robolectric to be able to test AsyncTask API.

👍 

> It also adds support for measuring codecoverage via kover plugin, as well as integration with Codecov: https://app.codecov.io/github/Parsely/parsely-android

Awesome, thanks for that and for trying [kotlinx-kover](https://github.com/Kotlin/kotlinx-kover)! 🥇 

> Kotlin Gradle Plugin is added as I plan to write tests in Kotlin and it was established internally, that we'll aim to rewrite the SDK to Kotlin as well.

👍 

-----

I have left few suggestions (💡) for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

-----

`EXTRAS`
1. Minor (🔍): The `testImplementation` dependencies got added as part of this 915e4e8eace81f92f4f841607301bd99904ebc8e commit. which was a bit off. I guess that was meant for that next a109e80918b9a32326c1f6cc809a6977825dd444 commit instead... 😅

parsely/build.gradle
wzieba: I removed the explicit declaration as it's not really needed - it defaults to JVM 1.8

settings.gradle
wzieba: Thanks for the suggestion! I have concerns with such change though: I don't see value in extracting values that are used only once (versions of all dependencies except `appcompat`). From my point of view, it only adds effort for anyone working with repository to look for version declaration, without any benefits.

About SDK versions: I don't think it's needed to add any more dependencies between `example` and `parsely` modules. They don't need to have the same version of SDK so I wouldn't introduce such policy.

parsely/build.gradle
ParaskP7: Got it, thanks for doing that! 👍

settings.gradle
ParaskP7: Thanks for taking a look @wzieba , for considering this suggestion and for raising some concerns! 🙇 

> I don't see value in extracting values that are used only once (versions of all dependencies except appcompat).

As you noticed, this is already helping with `appcompat`. IMHO, the main problem this is solving is not only any problems of today, but also problems of tomorrow where various dependencies will start having different version without any particular reason other than an oversight.

In addition to solving such problems, again, IMHO, this is making sure that all version are grouped and extracted in a central place where anyone could quickly take a look to understand what this project is using overall. Else, one would need to go and check that manually, module by module. Now, I understand that this project/repo has only 2 modules, with one of them being just a demo module, but still, that doesn't mean that this will not change in the future. Also, this doesn't mean that the demo module shouldn't be considered as important to use since this is our primarily means of testing any change.

Question (❓): If we were to use [Gradle Version Catalogs](https://docs.gradle.org/current/userguide/platforms.html) to share such versions universally and across every of our projects/repos (something already in progress), wouldn't that be the same? Would you be against doing that for the very same reasons, or is this somehow different? 🤔

> About SDK versions: I don't think it's needed to add any more dependencies between example and parsely modules. They don't need to have the same version of SDK so I wouldn't introduce such policy.

True, but at the same time, wouldn't it be better when a new SKD version is upgraded on the library, that the demo app is also being upgraded to verify that this could work there too, just in case? If we don't do that we could have one of the following:

1. The library module moving forward on any such SDK version, leaving the example module behind, the most likely scenario.
2. The example module moving forward on any such SDK version, leaving the library module behind, which to be honest, is less likely.
3. The library module is moving forward on any such SDK version, and when that is done, the engineer responsible for that change will also move forward on that SDK version on the example module as well, but now, this is done manually and probe to an oversight.

Wdyt? 🤔

wzieba: > As you noticed, this is already helping with appcompat.

I'm not sure if it really helps - I think there's nothing wrong with having two versions of a dependency if one of the projects is library and second is an example app.

> In addition to solving such problems, again, IMHO, this is making sure that all version are grouped and extracted in a central place where anyone could quickly take a look to understand what this project is using overall. Else, one would need to go and check that manually, module by module.

I'd argue that if someone wants to browse dependencies, they can use Gradle scan or `dependencies` task which shows all transitive dependencies - only then we can verify what dependencies are used really. Just browsing build files won't show accurate versions anyway.

> Question (❓): If we were to use [Gradle Version Catalogs](https://docs.gradle.org/current/userguide/platforms.html) to share such versions universally and across every of our projects/repos (something already in progress), wouldn't that be the same? Would you be against doing that for the very same reasons, or is this somehow different? 🤔

I don't think that such a small library will benefit from introducing Gradle Version Catalog. Also, the Gradle Version Catalog is different - it won't require from developers to extract and maintain versions directly in the repository. I just try to not add overhead, even minimal. Extracting things that are used only once, sounds like unnecessary work to me.

>True, but at the same time, wouldn't it be better when a new SKD version is upgraded on the library, that the demo app is also being upgraded to verify that this could work there too, just in case? If we don't do that we could have one of the following:

I'd be more concerned for the library targeting newer SDK version to work with older SDKs, so the scenario 1 or 3 sounds not as a bad option. If we want to really check different APIs, we should introduce tests for that.

ParaskP7: 👋 @wzieba and thanks for sharing some more thoughts on this, much appreciated!

I really value your opinion, and, don't want to change your mind on it, just sharing some more thoughts here, for completion purposes, in terms of argumentation on why I would always go in favor for such a change, thus always pushing on it, especially when the work required to do so is not too involved. 🙏 

> I'm not sure if it really helps - I think there's nothing wrong with having two versions of a dependency if one of the projects is library and second is an example app.

For sure, maybe for now and for a project as small as this it doesn't make much sense, but, to me, as long as it doesn't make it worse, I am always up to moving this forward and this way maybe even have some kind of consistency between all the projects/repos we are dealing with, at least in terms of dependency management.

Also, think about [this](https://github.com/Automattic/android-dependency-catalog) and how it can be used with this project/repo. Maybe even think Dependabot first, how would a tool like that, if integrated with such a project/repo, deal with two versions of a dependency. 🤔

> I'd argue that if someone wants to browse dependencies, they can use Gradle scan or dependencies task which shows all transitive dependencies - only then we can verify what dependencies are used really. Just browsing build files won't show accurate versions anyway.

Ah, I would love that, but this requires engineers to run a Gradle Scan first, I was thinking more on how to quick take a look at a project's dependencies at a quick glance (like it can be done [here](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/build.gradle#L17-L107)).

Also, doing something like that will make it easier for bots like Dependabot just update this one reference of the version instead of updating multiple references of such. 🤔

> I don't think that such a small library will benefit from introducing Gradle Version Catalog. Also, the Gradle Version Catalog is different - it won't require from developers to extract and maintain versions directly in the repository. I just try to not add overhead, even minimal.

As I said above, IMHO, as long as it doesn't make it worse, even if it will take a small amount of work, I am always up to moving this forward and this way maybe even have some kind of consistency between all the projects/repos we are dealing with, at least in terms of dependency management.

> Extracting things that are used only once, sounds like unnecessary work to me.

I understand that, but to me, the why of doing such work is all about consistency and readiness for future integrations.

> I'd be more concerned for the library targeting newer SDK version to work with older SDKs, so the scenario 1 or 3 sounds not as a bad option. If we want to really check different APIs, we should introduce tests for that.

👍

PR #80, opened 2023-10-17
Add Codecov token

wzieba:
Internal discussion about token generation: p1697550961218509-slack-C05C1C14UKF


ParaskP7 (approved):

wzieba:
Ah I see!

> Just to make sure of it, just for my information, we are using a separate CODECOV_TOKEN for these 2 projects, woocommerce-android and parsely-android, right? 🤔

Yes - those are two separate, completely unrelated tokens 👍 Thanks for the review!

PR #81, opened 2023-10-20
Add unit tests to `ParselyAPIConnection` and minor refactor of it

wzieba:
### Description

This PR adds unit tests for `ParselyAPIConnection` and adds some minor refactor to it.

### How to test

Review and running unit tests should be just fine. Due to async nature of, well, `AsyncTask`, you can also verify that each test execution waited for `AsyncTask` to finish. You should see `Pixel request success` or `Pixel request exception` at the end of test logs.

PR #82, opened 2023-10-24
Extract `EngagementManager` and cover it with unit tests

wzieba:
### Description

This PR extracts `EngagementManager` to a separate unit and adds unit tests for it. 

The proposed unit test is based on manual `Thread#sleep` calls and is time-sensitive. It's due to usage of `java.utils.Timer` which runs a new thread internally - a thread we cannot control. Still, the unit test is written in a way to not be flaky (usages of `org.assertj.core.api.isCloseTo` assertions) and brings value. E.g., the current test suite covers fix added in #62 .

### How to test

Review and green light on CI should be just fine.


parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Calendar;
+import java.util.TimeZone;
+
+class UpdateEngagementIntervalCalculator {
+
+    private static final long MAX_TIME_BETWEEN_HEARTBEATS = 60 * 60;
+    private static final long OFFSET_MATCHING_BASE_INTERVAL = 35;
+    private static final double BACKOFF_PROPORTION = 0.3;
+
+    long updateLatestInterval(@NonNull final Calendar startTime) {
```
wzieba, line 14: Unit tests for this class are planned to be added in a seperate PR.

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
```diff
@@ -0,0 +1,114 @@
...
+ * 1. Flushing any engaged time before canceling.
+ * 2. Progressive backoff for long engagements to save data.
+ */
+class EngagementManager {
+
+    private final ParselyTracker parselyTracker;
+    public Map<String, Object> baseEvent;
+    private boolean started;
+    private final Timer parentTimer;
+    private TimerTask waitingTimerTask;
```
ParaskP7, line 25: Suggestion (💡): I suggest making this `@Nullable` and then checking that during `stop()`:

```java
    public void stop() {
        if (waitingTimerTask != null) {
            waitingTimerTask.cancel();
        }
        started = false;
    }
```

PS: Actually, I would even recommend for such refactor/test PRs to add any missing nullability annotation wherever possible going forward. 🤷

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
```diff
@@ -0,0 +1,114 @@
...
+    private long latestDelayMillis, totalTime;
+    private Calendar startTime;
+    private final UpdateEngagementIntervalCalculator intervalCalculator;
+
+    public EngagementManager(ParselyTracker parselyTracker, Timer parentTimer, long intervalMillis, Map<String, Object> baseEvent, UpdateEngagementIntervalCalculator intervalCalculator) {
```
ParaskP7, line 30: Minor (🔍): This line exceed the usual `120` character limit guideline that we have on other projects, consider reformatting it to something more readable like:

```java
    public EngagementManager(
            ParselyTracker parselyTracker,
            Timer parentTimer,
            long intervalMillis,
            Map<String, Object> baseEvent,
            UpdateEngagementIntervalCalculator intervalCalculator
    ) {
```

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
```diff
@@ -0,0 +1,114 @@
...
+        started = true;
+        startTime = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+    }
+
+    public void stop() {
+        waitingTimerTask.cancel();
+        started = false;
+    }
+
+    public boolean isSameVideo(String url, String urlRef, ParselyVideoMetadata metadata) {
```
ParaskP7, line 55: Suggestion (💡): Consider resolving/suppressing all the warnings related to this method:

1. Suppress: `Unchecked cast: 'java.lang.Object' to 'java.util.Map<java.lang.String,java.lang.Object>'`
2. Resolve: `Method invocation 'equals' may produce 'NullPointerException'`
3. Resolve: `Unboxing of 'baseMetadata.get("duration")' may produce 'NullPointerException'`

FYI: For no.2 there is an easy fix, just flit the `equals(...)`, for example:

**BEFORE**: baseEvent.get("url").equals(url)
**AFTER**: url.equals(baseEvent.get("url")

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
```diff
@@ -0,0 +1,114 @@
...
+        // Create a copy of the base event to enqueue
+        Map<String, Object> event = new HashMap<>(baseEvent);
+        ParselyTracker.PLog(String.format("Enqueuing %s event.", event.get("action")));
+
+        // Update `ts` for the event since it's happening right now.
+        Calendar now = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+        @SuppressWarnings("unchecked")
+        Map<String, Object> baseEventData = (Map<String, Object>) event.get("data");
+        assert baseEventData != null;
+        Map<String, Object> data = new HashMap<>((Map<String, Object>) baseEventData);
```
ParaskP7, line 96: Minor (🔍): Consider removing the redundant `(Map<String, Object>)` casting for `baseEventData`.

Warning: `Casting 'baseEventData' to 'Map<String, Object>' is redundant`

parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
```diff
@@ -0,0 +1,23 @@
...
+        Calendar now = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+        long totalTrackedTime = (now.getTime().getTime() - startTime.getTime().getTime()) / 1000;
+        double totalWithOffset = totalTrackedTime + OFFSET_MATCHING_BASE_INTERVAL;
+        double newInterval = totalWithOffset * BACKOFF_PROPORTION;
+        long clampedNewInterval = (long) Math.min(MAX_TIME_BETWEEN_HEARTBEATS, newInterval);
+        System.out.println("New interval: " + clampedNewInterval*1000);
```
ParaskP7, line 20: 1. Warning (⚠️): I am not sure if this was done due to an oversight or by design but usually `ParselyTracker.PLog(...)` is used instead. Consider either removing this completely or using `ParselyTracker.PLog(...)` instead.
2. Question (❓): Now that I am seeing the `System.out.println(...)` logging, what about replacing the whole `ParselyTracker.PLog(...)` with the native Android logger, of course, and this way to be having a more fine-grained level of logging (`Log.d/i/e(...)`), as a separate issue/PR, wdyt? 🤔

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+import org.assertj.core.api.MapAssert
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+private typealias Event = MutableMap<String, Any>
+
+@RunWith(RobolectricTestRunner::class)
+@Suppress("UNCHECKED_CAST")
```
ParaskP7, line 20: Minor (🔍): Consider moving this suppression closer to the warning, that is, instead of having it here for the class class, to only apply that to the `isCorrectEvent(...)` extension function.

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    class FakeIntervalCalculator : UpdateEngagementIntervalCalculator() {
+        override fun updateLatestInterval(startTime: Calendar): Long {
+            return DEFAULT_INTERVAL_MILLIS
+        }
+    }
+
+    private val now: Long
```
ParaskP7, line 141: Minor (🔍): Since this is a `val` field and not a method or an inner class, for better visibility, consider moving that at the top, alongside the rest of the fields.

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        get() = Calendar.getInstance(TimeZone.getTimeZone("UTC")).timeInMillis
+
+    private fun sleep(millis: Long) = Thread.sleep(millis + THREAD_SLEEPING_THRESHOLD)
```
ParaskP7, line 144: Minor (🔍): Since this is a `private` function and not an inner class, for better visibility, consider moving that amongst the other private functions.

FYI: I would have moved that above the `isCorrectEvent(...)` extension method since it is always being used first to that during testing.

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    @Test
+    fun `when starting manager, then record the correct event after interval millis`() {
+        // when
+        sut.start()
+        sleep(DEFAULT_INTERVAL_MILLIS)
+        val timestamp = now - THREAD_SLEEPING_THRESHOLD
+
+        // then
+        assertThat(tracker.events[0]).isCorrectEvent(
```
ParaskP7, line 50: Minor (🔍): Consider making this assertion more readable, just like you did with the test further below:

```kotlin
        assertThat(tracker.events[0]).isCorrectEvent(
            // Ideally: totalTime should be equal to DEFAULT_INTERVAL_MILLIS
            withTotalTime = { isCloseTo(DEFAULT_INTERVAL_MILLIS, withinPercentage(10)) },
            // Ideally: timestamp should be equal to System.currentTimeMillis() at the time of recording the event
            withTimestamp = { isCloseTo(timestamp, within(100L)) }
        )
```

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+            withTotalTime = {
+                // Ideally: totalTime should be equal to DEFAULT_INTERVAL_MILLIS
+                isCloseTo(DEFAULT_INTERVAL_MILLIS, withinPercentage(10))
+            },
+            withTimestamp = {
+                // Ideally: timestamp should be equal to System.currentTimeMillis() at the time of recording the event
+                isCloseTo(
+                    timestamp,
+                    within(100L)
```
ParaskP7, line 59: Minor (🔍): On the tests below you are using `20L` instead of `100L`. Consider changing that value to `20L` as well.

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+                )
+            }
+        )
+    }
+
+    @Test
+    fun `when starting manager, then schedule task each interval period`() {
```
ParaskP7, line 66: Question (❓): Please help me understand how `start()` works on `EngagementManager` for multiple events. 🙏

FYI: I am asking that because, to my understanding, a new instance of `EngagementManager` is created with just one event, the `baseEvent`, and that is the only event needed to be enqueued. I am struggling a bit to understand the logic here and when we can have multiple of the same events queuing. 🤔

PS: This is related to 51a467463c5f0602100a436f76d21bed2f375db9 .

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+            // Ideally: timestamp should be equal to `now` at the time of recording the event
+            withTimestamp = { isCloseTo(thirdTimestamp, within(20L)) }
+        )
+    }
+
+    private fun MapAssert<String, Any>.isCorrectEvent(
+        withTotalTime: AbstractLongAssert<*>.() -> AbstractLongAssert<*>,
+        withTimestamp: AbstractLongAssert<*>.() -> AbstractLongAssert<*>,
+    ): MapAssert<String, Any> {
+        return containsEntry("action", "heartbeat")
```
ParaskP7, line 107: Minor (🔍): Consider moving the `.hasEntrySatisfying("data")` check above the `.containsEntry("inc", 0L)` check to make it easier to follow the same logical order found within the `EngagementManager.doEnqueue(...)` method, that is, without having to jump up and down to verify test correctness.

```kotlin
        return containsEntry("action", "heartbeat")
            // Incremental will be always 0 because the interval is lower than 1s
            .hasEntrySatisfying("data") { data ->
                data as Map<String, Any>
                assertThat(data).hasEntrySatisfying("ts") { timestamp ->
                    timestamp as Long
                    assertThat(timestamp).withTimestamp()
                }.containsAllEntriesOf(testData.minus("ts"))
            }.containsEntry("inc", 0L)
            .hasEntrySatisfying("tt") { totalTime ->
                totalTime as Long
                assertThat(totalTime).withTotalTime()
            }
```

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        const val THREAD_SLEEPING_THRESHOLD = 50L
+        val testData = mutableMapOf<String, Any>(
+            "os" to "android",
+            "parsely_site_uuid" to "e8857cbe-5ace-44f4-a85e-7e7475f675c5",
+            "os_version" to "34",
+            "manufacturer" to "Google",
+            "ts" to 123L
+        )
+    }
+}
```
ParaskP7, line 158: Minor (🔍): Consider adding an empty line at the end of this test.

FYI: You can enable the `Ensure every saved file ends with a line break` option within your IDE.

PS: You will find that within `Editor` -> `General` -> `On Save` section at the bottom.

ParaskP7 (changes requested):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, another great testing job well done! 🌟 🧪 🌟 

But wow, that is a very convoluted test, I am very sorry you had to go through all this... 🫂 

-----

I have left a warning (⚠️). a couple of questions (❓), a few suggestions (💡) and some minor (🔍) comments for you to consider. I am going to NOT going to approve this PR for now, just because of this one warning (⚠️) comment, but other than that everything seems okay to me.

-----

`EXTRAS`
1. Suggestion (💡): I have noticed that during you refactor commits some of them wouldn't be able to build, just because it would require this other subsequent commit that you add later on. I would personally merge those 2 commits together and make it easier for the reviewer to understand that in order to (for example) extract a class, one would also need to make some methods more public (package private). 🤔
2. Suggestion (💡): In this PR, I am also seeing some additional commits that are not part of this PR. Consider rebasing you branch with the `main` one before open a PR for review. This will make it easier for reviewers to know exactly what need to be reviewed and avoid having them to figure out the out-of-scope commits.

parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
```diff
@@ -0,0 +1,23 @@
+package com.parsely.parselyandroid;
+
+import androidx.annotation.NonNull;
+
+import java.util.Calendar;
+import java.util.TimeZone;
+
+class UpdateEngagementIntervalCalculator {
```
ParaskP7, line 8: Suggestion (💡): Ah, I forgot to mention this in my review, what about making this a `Utils` class instead of instantiating it as an object. Maybe even a Kotlin utils class, wdyt? 🤔

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
wzieba: Sure, but I'd leave this, and other static code changes, to some linter. I'd prefer to not focus on this on reviews, as I'm afraid it'll shadow other, more functional comments. Do you have a suggestion for a linter/formatter tool for Java?

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
wzieba: I'd leave this to a different PR - I wanted this PR to be focused on extracting `EngagementManager` and adding minimal set of changes to make it testable.

parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
wzieba: Thanks, I'll remove this log.

> System.out.println(...) logging, what about replacing the whole ParselyTracker.PLog(...) with the native Android logger,

We could do this, but at the same time, I don't see really a value in such change for this SDK. Could you elaborate, what would we gain by using the Android logger?

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
wzieba: When we call `start` on `EngagementManager`, the `java.utils.TimerTask` is scheduled inside `java.utils.Timer`. The next execution of the `TimerTask` will schedule a subsequent task.

So this test checks if subsequent tasks are executed and if they have correct parameters like e.g. updated total timestamp. Does this answer your questions @ParaskP7 ?

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
ParaskP7: > Sure, but I'd leave this, and other static code changes, to some linter. I'd prefer to not focus on this on reviews, as I'm afraid it'll shadow other, more functional comments.

Sure, this is just a minor one, feel free to skip those when needed. 👍 

> Do you have a suggestion for a linter/formatter tool for Java?

On most of our other Java-utilizing repos, we are and have been using `CheckStyle`, but then again we are not actively updating this tool on a regular basis, or at all for that matter, thus I am not sure. Having said that, `CheckStyle` indeed seems to be the best we can do for now, in only for consistency purpose... 🤷

parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
ParaskP7: > Could you elaborate, what would we gain by using the Android logger?

Well, as I said above, a more fine-grained level of logging (`Log.d/i/e(...)`). Apart from that, I am not sure sure if using `System.out.println(...)` logging has some performance implication for apps that are using this SDK, or even if apps can remove this logging if they want to, with the native Android logging, I think the can, right... 🤷 🤔

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
wzieba: If that's okay, I'd leave this as it is. My reasoning: on top, we have declarations for the test subject and its dependencies. At the bottom: fakes and helper methods for test execution. `now` being `val` is only a syntax sugar to not add `()` in some places.

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
ParaskP7: Hmmm, yes, I understand how the `TimerTask` will schedule a subsequent task. I just don't understand when multiple events, mostly in production, could be scheduled, can you help me with that? 🙏 

```java
        engagementManager = new EngagementManager(this, timer, DEFAULT_ENGAGEMENT_INTERVAL_MILLIS, event, intervalCalculator);
        engagementManager.start();
```

I see the above happening on production, which on (for example) `startEngagement(...)` creates a new instance of `EngagementManager` with a specific single `event`. How is this manager is able to handle multiple events, when does this happens, this is what I don't understand @wzieba . 🤔

parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
wzieba: Gotcha, thanks. I've added an issue for that: https://github.com/Parsely/parsely-android/issues/83 - I thought that `PLog` logs only if SDK has `debug` flag to `true`.

ParaskP7: Awesome, thanks for opening this issue @wzieba ! 🙇 

> I thought that `PLog` logs only if SDK has `debug` flag to `true`.

Wishful thinking... 😅 🤷

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
wzieba: >  How is this manager is able to handle multiple events, when does this happens, this is what I don't understand @wzieba . 🤔

When you say "multiple events" do you mean "multiple engagement sessions"? If yes, then the SDK cannot handle it.

Circling back to the original question:

> I am asking that because, to my understanding, a new instance of EngagementManager is created with just one event, the baseEvent, and that is the only event needed to be enqueued.

Yes, the `EngegementManager` is created with just one event, the `baseEvent`. But no - it's not the only event needed to be enqueued. After the initial `baseEvent` is enqueued, enqueuing of the next event is scheduled.

When interval time passes, the SDK, based on `baseEvent`, enqueues a new event with modifications done in `doEnqueue`. It modifies, e.g. `ts`, `inc` or `tt` parameters.

ParaskP7: > When you say "multiple events" do you mean "multiple engagement sessions"? If yes, then the SDK cannot handle it.

Good to know! 💯 

> Yes, the EngegementManager is created with just one event, the baseEvent. But no - it's not the only event needed to be enqueued. After the initial baseEvent is enqueued, enqueuing of the next event is scheduled.

👍 

> When interval time passes, the SDK, based on baseEvent, enqueues a new event with modifications done in doEnqueue. It modifies, e.g. ts, inc or tt parameters.

That explains it, and apologies for only knowing the basics of the SDK, trying to figure out the rest during reviews... 😊 

So, when `startEngagement(...)` gets triggered, it is always triggered with a `baseEvent`, and then, until `stopEngagement(...)` is called, the same `baseEvent` will be enqueued multiple time, based on the interval, with only some of its parameters being modified. If so, me repeating correctly what you already said, I then think I got the complete picture now, thanks so much for explaining. 🙇

parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
wzieba: In next iterations, I'll inject some `Clock` to this class to fake the current time. Otherwise, test for this class will be time-sensitive. So I'll have to leave this as a class, not a static one.

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
wzieba: I've fixed the comments! After the second thought, it's probably not worth to introduce Java linter - we plan to use Kotlin anyway.

parsely/src/main/java/com/parsely/parselyandroid/UpdateEngagementIntervalCalculator.java
ParaskP7: Nice to know, thanks for the heads-up on that. 💯

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.java
ParaskP7: > I've fixed the comments! 

Checking @wzieba and thanks! 👀 

> After the second thought, it's probably not worth to introduce Java linter - we plan to use Kotlin anyway.

Yeaaa, maybe not... 🤷

ParaskP7 (approved):
Took it for another spin, everything LGTM now @wzieba , thanks again! 🚀

parsely/src/test/java/com/parsely/parselyandroid/EngagementManagerTest.kt
wzieba: > the same baseEvent will be enqueued multiple time, based on the interval, with only some of its parameters being modified

Yes, exactly this 🙌 !

wzieba:
Thank you @ParaskP7 ! 🎉

wzieba:
@ParaskP7 one question on the initial review:

> Suggestion (💡): In this PR, I am also seeing some additional commits that are not part of this PR. Consider rebasing you branch with the main one before open a PR for review.

Which additional commits did you have in mind?

ParaskP7:
👋 @wzieba and thanks for the reply! 🙇 

Those [two below](https://github.com/Parsely/parsely-android/pull/82/commits) got me a bit confused:

image

🤷 

I remembered reviewing those two on this #81 PR, see bottom of this [commit list](https://github.com/Parsely/parsely-android/pull/81/commits):

image

🤷 

Does that help? 🙏

wzieba:
I see, thanks! It's interesting though that those commits have different hashes. Normally, we shouldn't see those commits after merging `main`.

ParaskP7:
Yeaaa, not sure what happened there, that's why, by default, I always recommend rebasing branches with latest `main` branch before opening a PR... 🤷

PR #84, opened 2023-10-26
Fix heartbeats interval calculations, add unit tests for the calculator, rewrite to Kotlin

wzieba:
### Description

This PR fixes the minor off difference when calculating heartbeat intervals (magnitude of up to 900ms). I won't change anything drastically, yet it's better to be in sync with iOS and have a clear implementation, without "hidden casting" issues.

It also covers the calculator with unit tests and changes the calculator implementation to use less error-prone `kotlin.time.Duration` class to make arithmetical operations between different times.

iOS implementation of this behavior: https://github.com/Parsely/AnalyticsSDK-iOS/blob/b1db4b05dcd5af6feb4710ec30d13fa704305692/Sources/Sampler.swift#L136


parsely/src/main/java/com/parsely/parselyandroid/HeartbeatIntervalCalculator.kt
```diff
@@ -0,0 +1,27 @@
+package com.parsely.parselyandroid
+
+import java.util.Calendar
+import kotlin.time.Duration.Companion.hours
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+internal open class HeartbeatIntervalCalculator(private val clock: Clock) {
```
ParaskP7, line 9: 1. Praise (❤️): Great work on this, I love how this class is now using duration and the `x.seconds/minutes/hours` approach, much more readable. Also, kudos for fixing the bugs of unwanted rounding (f273ee142c647cfc6a6ee24b3d68bf98acc2734d)! 🥇  

> It also covers the calculator with unit tests and changes the calculator implementation to use less error-prone kotlin.time.Duration class to make arithmetical operations between different times.

3. Suggestion (💡): When converting a Java class to Kotlin (98238156e20ae7b2bff6249a8d18af09f04dbd94), consider using the `Extra commit for .java > .kt renames` option while committing this change. This will make it easier for anyone reviewing this change later-on as it will show an actual diffing of files instead of showing the Java file being deleted and a new Kotlin file being added.

image

FYI: For more context on check [this discussion](https://github.com/wordpress-mobile/WordPress-Android/issues/17260#issuecomment-1549630646) I had with an external contributor a while back. Let me know your thoughts. 🙏

parsely/src/main/java/com/parsely/parselyandroid/HeartbeatIntervalCalculator.kt
```diff
@@ -0,0 +1,27 @@
...
+        val totalWithOffset = totalTrackedTime + OFFSET_MATCHING_BASE_INTERVAL
+        val newInterval = totalWithOffset * BACKOFF_PROPORTION
+        val clampedNewInterval = minOf(MAX_TIME_BETWEEN_HEARTBEATS, newInterval)
+        return clampedNewInterval.inWholeMilliseconds
+    }
+
+    companion object {
+        const val BACKOFF_PROPORTION = 0.3
+        val OFFSET_MATCHING_BASE_INTERVAL = 35.seconds
+        val MAX_TIME_BETWEEN_HEARTBEATS = 15.minutes
```
ParaskP7, line 25: > This PR fixes the minor off difference when calculating heartbeat intervals (magnitude of up to 900ms). I won't change anything drastically, yet it's better to be in sync with iOS and have a clear implementation, without "hidden casting" issues.

👍

parsely/src/test/java/com/parsely/parselyandroid/HeartbeatIntervalCalculatorTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        val result = sut.calculate(startTime)
+
+        // then
+        // ((currentTime + offset) * backoff) and then in milliseconds
+        // (2 + 35) * 0.3 * 1000 = 11100
+        assertThat(result).isEqualTo(11100)
+    }
+
+    class FakeClock : Clock() {
+        var fakeNow = Duration.ZERO
```
ParaskP7, line 77: Praise (❤️): Nice!

parsely/src/test/java/com/parsely/parselyandroid/HeartbeatIntervalCalculatorTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    fun `given a time that will cause the interval to surpass the MAX_TIME_BETWEEN_HEARTBEATS, when calculating interval, then return the MAX_TIME_BETWEEN_HEARTBEATS`() {
+        // given
+        // "excessiveTime" is a calculated point in time where the resulting interval would
+        // surpass MAX_TIME_BETWEEN_HEARTBEATS
+        // (currentTime + offset) * backoff = max
+        // currentTime = (max / backoff) - offset, so
+        // (15 minutes / 0.3) - 35 seconds = 2965 seconds. Add 1 second to be over the limit
```
ParaskP7, line 45: Praise (❤️): Thanks for adding these comments and making it clear how these numbers are calculated. 🙇

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, great job + Kotlin in production! 🌟 🎉 🌟

parsely/src/main/java/com/parsely/parselyandroid/HeartbeatIntervalCalculator.kt
wzieba: Thanks for sharing about `Extra commit` option, TIL!

PR #85, opened 2023-10-29
Add setup for functional tests. Test sending events over the local variable limit.

wzieba:
## Description

This PR adds a setup for writing and running functional tests on CI. Contrary to unit tests, they are meant to be written via black-box technique. The only mocked/faked element of the setup is API, which is intentionally mocked via MockWebServer.

Additionally, it adds a test case for tracking number of events exceeding maximum number of events that are stored in memory. This asserts correctness of saving/querying data from the local file as well as shows the benefits of functional tests.

### Note

Important pivot how to approach the proposed test was made in 8d4a0d39e04b57a6a05bb260e6978e23905c7c12 so please don't pay a lot of attention to implementation details introduced in commit before, which is e97a7a5c3c9ffd2bf75b6b986969e231e021a09a.

## How to test

Green light on CI should be enough. You can also download the build artifact and see results of instrumentation tests.

### DIFF
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -4,6 +4,11 @@ plugins {
     id 'org.jetbrains.kotlinx.kover'
 }
 
+ext {
+    assertJVersion = '3.24.2'
+    mockWebServerVersion = '4.12.0'
+}
+
 android {
     compileSdkVersion 33
 
@@ -59,17 +64,17 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.3'
     implementation 'com.google.android.gms:play-services-ads-identifier:18.0.1'
     implementation 'androidx.lifecycle:lifecycle-process:2.6.2'
-    
+
     testImplementation 'org.robolectric:robolectric:4.10.3'
     testImplementation 'androidx.test:core:1.5.0'
-    testImplementation 'org.assertj:assertj-core:3.24.2'
+    testImplementation "org.assertj:assertj-core:$assertJVersion"
     testImplementation 'junit:junit:4.13.2'
-    testImplementation 'com.squareup.okhttp3:mockwebserver:4.12.0'
+    testImplementation "com.squareup.okhttp3:mockwebserver:$mockWebServerVersion"
 
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test:rules:1.5.0'
-    androidTestImplementation 'org.assertj:assertj-core:3.24.2'
-    androidTestImplementation 'com.squareup.okhttp3:mockwebserver:4.12.0'
+    androidTestImplementation "org.assertj:assertj-core:$assertJVersion"
+    androidTestImplementation "com.squareup.okhttp3:mockwebserver:$mockWebServerVersion"
     androidTestImplementation 'androidx.test:runner:1.5.2'
     androidTestUtil 'androidx.test:orchestrator:1.4.2'
 }
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -10,6 +10,8 @@ android {
     defaultConfig {
         minSdkVersion 21
         targetSdkVersion 33
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
     buildTypes {
         release {
@@ -22,6 +24,7 @@ android {
         exclude 'META-INF/LICENSE'
         exclude 'META-INF/NOTICE'
     }
+    testBuildType "debug"
 
     publishing {
         singleVariant('release') {
@@ -58,6 +61,12 @@ dependencies {
     testImplementation 'org.assertj:assertj-core:3.24.2'
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'com.squareup.okhttp3:mockwebserver:4.12.0'
+
+    androidTestImplementation 'androidx.test.ext:junit:1.1.5'
+    androidTestImplementation 'androidx.test:rules:1.5.0'
+    androidTestImplementation 'org.assertj:assertj-core:3.24.2'
+    androidTestImplementation 'com.squareup.okhttp3:mockwebserver:4.12.0'
+    androidTestImplementation 'androidx.test:runner:1.5.2'
 }
 
 apply from: "${rootProject.projectDir}/publication.gradle"
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -33,7 +33,6 @@ android {
         exclude 'META-INF/LICENSE'
         exclude 'META-INF/NOTICE'
     }
-    testBuildType "debug"
 
     publishing {
         singleVariant('release') {
--- a/parsely/build.gradle
+++ b/parsely/build.gradle
@@ -12,6 +12,10 @@ android {
         targetSdkVersion 33
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        testInstrumentationRunnerArguments clearPackageData: 'true'
+    }
+    testOptions {
+        execution 'ANDROIDX_TEST_ORCHESTRATOR'
     }
     buildTypes {
         release {
@@ -67,6 +71,7 @@ dependencies {
     androidTestImplementation 'org.assertj:assertj-core:3.24.2'
     androidTestImplementation 'com.squareup.okhttp3:mockwebserver:4.12.0'
     androidTestImplementation 'androidx.test:runner:1.5.2'
+    androidTestUtil 'androidx.test:orchestrator:1.4.2'
 }
 
 apply from: "${rootProject.projectDir}/publication.gradle"
--- a/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
+++ b/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
@@ -1,0 +1,24 @@
+package com.parsely.parselyandroid
+
+import android.app.Activity
+import androidx.test.core.app.ActivityScenario
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Test
+import org.junit.runner.RunWith
+
+
+@RunWith(AndroidJUnit4::class)
+class FunctionalTests {
+
+    private lateinit var parselyTracker: ParselyTracker
+
+    @Test
+    fun appTracksEventsAboveQueueSizeLimit() {
+        ActivityScenario.launch(SampleActivity::class.java).use { scenario ->
+            scenario.onActivity { activity: Activity ->
+            }
+        }
+    }
+
+    class SampleActivity : Activity()
+}
--- a/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
+++ b/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
@@ -22,7 +22,6 @@ import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import okhttp3.mockwebserver.RecordedRequest
 import org.assertj.core.api.Assertions.assertThat
-import org.assertj.core.api.Assertions.fail
 import org.junit.Test
 import org.junit.runner.RunWith
 
@@ -65,16 +64,9 @@ class FunctionalTests {
             val requestPayload = server.takeRequest().toMap()
             assertThat(requestPayload["events"]).hasSize(51)
 
-            runBlocking {
-                withTimeoutOrNull(500.milliseconds) {
-                    while (true) {
-                        yield()
-                        if (locallyStoredEvents.size == 0) {
-                            break
-                        }
-                    }
-                } ?: fail("Local storage file is not empty!")
-            }
+            // Wait a moment to give SDK time to delete the content of local storage file
+            waitFor { locallyStoredEvents.isEmpty() }
+            assertThat(locallyStoredEvents).isEmpty()
         }
     }
 
@@ -86,7 +78,7 @@ class FunctionalTests {
     }
 
     @JsonIgnoreProperties(ignoreUnknown = true)
-    data class Event (
+    data class Event(
         @JsonProperty("idsite") var idsite: String,
     )
 
@@ -115,4 +107,15 @@ class FunctionalTests {
     }
 
     class SampleActivity : Activity()
+
+    private fun waitFor(condition: () -> Boolean) = runBlocking {
+        withTimeoutOrNull(500.milliseconds) {
+            while (true) {
+                yield()
+                if (condition()) {
+                    break
+                }
+            }
+        }
+    }
 }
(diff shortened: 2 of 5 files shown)


parsely/build.gradle
```diff
@@ -4,12 +4,23 @@ plugins {
...
 
     defaultConfig {
         minSdkVersion 21
         targetSdkVersion 33
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        testInstrumentationRunnerArguments clearPackageData: 'true'
+    }
+    testOptions {
+        execution 'ANDROIDX_TEST_ORCHESTRATOR'
```
ParaskP7, line 23: Praise (❤️): TIL about that and the fact that this makes sure that this local storage file won't be shared between test (56c0fe864b332feaf7eae49894b06d8c232b3f95), and if removed, this will make the test to fail as the local storage file name will now exist, interesante! 🧪

parsely/build.gradle
```diff
@@ -22,6 +33,7 @@ android {
         exclude 'META-INF/LICENSE'
         exclude 'META-INF/NOTICE'
     }
+    testBuildType "debug"
```
ParaskP7, line 36: Question (❓): I noticed this, plus the fact that you are using `./gradlew :parsely:connectedCheck` on CI for the functional tests, why is that? I mean, why not doing exactly what've done for the unit test, using the `Debug` variant for the task name, like `connectedDebugAndroidTest` as using the `connectedCheck` is a super task that incorporate more checks than just running the Android instrumentation tests, which is what we want here, right?

FYI: This way, removing this constraint, it will also enable us to use the `release` build type in the future, should we need it for Android tests, right? 🤔

PS: This is just an FYI question that I want to understand more, not to question it's validity, I am sure there is a good reason you used `connectedCheck` here. 🙏

parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
```diff
@@ -0,0 +1,121 @@
+package com.parsely.parselyandroid
+
+import android.app.Activity
+import androidx.test.core.app.ActivityScenario
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.core.type.TypeReference
+import com.fasterxml.jackson.databind.ObjectMapper
+import java.io.File
```
ParaskP7, line 10: Minor (🔍): For some reason, when I run reformat on this file it moved those imports at the end, I am not sure if you have a different imports configuration on your side, if that's the default or not, but maybe we can align on that so that we don't get a diff between us. 🤔

parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
```diff
@@ -0,0 +1,121 @@
...
+
+                repeat(51) {
+                    parselyTracker.trackPageview("url", null, null, null)
+                }
+            }
+
+            // Waits for the SDK to send events (flush interval passes)
+            val requestPayload = server.takeRequest().toMap()
+            assertThat(requestPayload["events"]).hasSize(51)
+
```
ParaskP7, line 66: Question (❓): Although I understand how this test is working and what it tried to verify, I feel that something is missing here, an assertion of some kind, something that will showcase what was documented on the test case above:

- `The SDK will save the events to disk...`
- `...and send them in the next flush interval.`
- `At the end, when all events are sent...`
- `...the SDK will delete the content of local storage file.`

I mean, can we assert that the SDK is saving the events first? For example, will it save all `51` of them, or only the `1` event that exceed the threshold, something else? Then, what happens during the first flush interval, what happens during the next flush interval, these kind of question. Apologies for trying to push things, I am just trying to review this as a very naive test reader that would like to understand this SDK through testing, assuming I don't have any prior knowledge or the SDK. Wdyt, too much? 🤔

PS: I also naively tried to add something basic, the `assertThat(locallyStoredEvents).isNotEmpty` before the `waitFor { locallyStoredEvents.isEmpty() }`, to see if that will pass, but it failed. 🤷

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
```diff
@@ -61,7 +61,10 @@ public class ParselyTracker {
     private static final String STORAGE_KEY = "parsely-events.ser";
 // emulator localhost
 //    private static final String ROOT_URL = "http://10.0.2.2:5001/";
-    private static final String ROOT_URL = "https://p1.parsely.com/";
+    /**
+     * @noinspection StringOperationCanBeSimplified
+     */
+    private static final String ROOT_URL = "https://p1.parsely.com/".intern();
```
ParaskP7, line 67: 1. Praise (❤️): TIL about that and `intern()`, never seen that before (ed56a27b99ef0c45cdcd466516f089b780cf910a)! 😮
2. Question (❓): I wonder whether we can do something better about this `ROOT_URL`, it feels to be better suited for a configuration file and not something that should exist as a private static final string... 🤷 Although, I do understand that configuring this should be possible as we always want to point to this url, as such I am a bit torn here... 🤔
3. Suggestion (💡): I suggest using `@SuppressWarnings("StringOperationCanBeSimplified")` instead of suppressing that via a Javadoc, just like we do elsewhere.
4. Minor (🔍): I suggest adding `.intern()` on the emulator localhost related `ROOT_URL` as well, just to avoid any question on why one is with and why the other without... 🤷 
5. Suggestion (💡): Also, as I recommended before, I again suggest refactoring this production `ROOT_URL` and its emulator equivalent to become side-by-side so that when commenting that out-and-in it will become as effortless as possible, with the suppress warning still being applicable to both before-and-after the change:

`ROOT_URL` production:

image

`ROOT_URL` emulator localhost:

image

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, another great job, more testing, more loving! 🌟 🧪 ❤️ 

-----

I have left few questions (❓), a couple of suggestions (💡) and some minor (🔍) comments for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
wzieba: > I mean, can we assert that the SDK is saving the events first?

This is something I pivoted from in 8d4a0d39e04b57a6a05bb260e6978e23905c7c12. IMO, the functional tests shouldn't check for such implementation details. The things we're interested in/testing here are:

- that SDK has sent the same number of events consumer app tracked (asserted in line 65)
- that SDK doesn't store additional data (line 69). Actually, we shouldn't even check this probably in a functional test, but this is a safe-check as I know we have some issues with events duplications.

> Then, what happens during the first flush interval, what happens during the next flush interval, these kind of question.

There is only one flush, and it's tested 👍 I'll be extending this tests suite to cover more cases, like multiple flushes, but here we test "what would happen if user recorded more than 50 events quickly".

> PS: I also naively tried to add something basic, the assertThat(locallyStoredEvents).isNotEmpty before the waitFor { locallyStoredEvents.isEmpty() }, to see if that will pass, but it failed. 🤷

Good point, but it's only timing thing. It seems that assertions above take long enough to give SDK time to clean the file.

Try this:

```diff
Subject: [PATCH] refactor: move the class to Kotlin

This class is relatively simple and difficult to test because of `java.utils.Timer`. That's why I decided to make migration to Kotlin right away, without unit tests coverage first.
---
Index: parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt b/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
--- a/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt	(revision 5900d93eff388bc5b2d98c44072cee07d91b1a96)
+++ b/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt	(date 1698666962291)
@@ -62,11 +62,11 @@
 
             // Waits for the SDK to send events (flush interval passes)
             val requestPayload = server.takeRequest().toMap()
+            assertThat(locallyStoredEvents).isNotEmpty
             assertThat(requestPayload["events"]).hasSize(51)
 
             // Wait a moment to give SDK time to delete the content of local storage file
             waitFor { locallyStoredEvents.isEmpty() }
-            assertThat(locallyStoredEvents).isEmpty()
         }
     }
 ```

and the test should fail!

WDYT?

File: parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt (hunk 3 above)
Comment (wzieba) [reply parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt:10]: I don't know, maybe: I use "Default IDE" in this project. I think it's okay, as it's sorted alphabetically.

Comment (ParaskP7) [reply parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt:10]: Yea, I think the default IDE is per the scheme, which applies to all projects, not a specific project (🤷). Maybe we should compare our default and update them on our imports, either you or me, both to have us both on the same page...

File: parsely/build.gradle (hunk 2 above)
Comment (wzieba) [reply parsely/build.gradle:36]: Good point, thanks! Actually, there's no really a reason why I used `connectedCheck`: fixed this in https://github.com/Parsely/parsely-android/pull/85/commits/d2a25dbc864396d67dc517943a5a079cffb86304 🙂 👍

File: parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt (hunk 3 above)
Comment (wzieba) [reply parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt:10]: Oh I see now! I think about introducing `detekt` to this project soon, so probably we won't need to sync the settings - the `detekt` default config will do that for us. WDYT?

File: parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt (hunk 4 above)
Comment (ParaskP7) [reply parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt:66]: Thanks for the reply @wzieba ! 👍 

> This is something I pivoted from in https://github.com/Parsely/parsely-android/commit/8d4a0d39e04b57a6a05bb260e6978e23905c7c12. IMO, the functional tests shouldn't check for such implementation details. The things we're interested in/testing here are:
> 
> - that SDK has sent the same number of events consumer app tracked (asserted in line 65)
> - that SDK doesn't store additional data (line 69). Actually, we shouldn't even check this probably in a functional test, but this is a safe-check as I know we have some issues with events duplications.

👍 if you are also 👍 with that, I just wanted to question it a bit... 😊

> There is only one flush, and it's tested 👍 I'll be extending this tests suite to cover more cases, like multiple flushes, but here we test "what would happen if user recorded more than 50 events quickly".

Thanks for the clarification on that. 👍 

> Good point, but it's only timing thing. It seems that assertions above take long enough to give SDK time to clean the file.
> 
> Try this:
> 
> + assertThat(locallyStoredEvents).isNotEmpty
> 
> and the test should fail!
> 
> WDYT?

Yea, it is hard to assert those things... 😭 

FYI: The test should fail, I don't understand that, the test should success with this change, is it not? It failing was the actual problem I had as well, as I would have expected it to succeed before waiting... 🤔

PS: Apologies, I am a bit confused it seems... 🤷 

> - assertThat(locallyStoredEvents).isEmpty()

Why did you remove this assertion from the patch you shared? 🤔

File: parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt (hunk 3 above)
Comment (ParaskP7) [reply parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt:10]: Yea, let's forget about this until `Detekt` is added to this project, deal! 👍

File: parsely/build.gradle (hunk 2 above)
Comment (ParaskP7) [reply parsely/build.gradle:36]: Awesome, thanks for this change!

Btw, now that you did d2a25dbc864396d67dc517943a5a079cffb86304 , do we actually need this `testBuildType "debug"` configuration? 🤔

File: parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java (hunk 5 above)
Comment (ParaskP7) [reply parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java:64]: 5dcb5ffb5e8db8a8fe373c7874816529f4b44b9e + a266ab00eacdb003f6d282bbc80b294de556c797 is looking good, thanks! 👍

File: parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt (hunk 4 above)
Comment (wzieba) [reply parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt:66]: Yes, sorry I made some confusion with the diff! Please ignore the diff in previous comment. But it's still case of timing - check this:
```diff
Subject: [PATCH] fix: add `intern` to `ROOT_URL` of emulator localhost
---
Index: parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt b/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
--- a/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt	(revision d2a25dbc864396d67dc517943a5a079cffb86304)
+++ b/parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt	(date 1698668794333)
@@ -71,6 +71,7 @@
     }
 
     private fun RecordedRequest.toMap(): Map<String, List<Event>> {
+        assertThat(locallyStoredEvents).isEmpty()
         val listType: TypeReference<Map<String, List<Event>>> =
             object : TypeReference<Map<String, List<Event>>>() {}
 

```

In this case the test **will** fail although, in theory, it shouldn't because after making the request, SDK will clean the local file. But, as the things happen on different threads, we don't have certainty what will happen first - assertion, or file deletion.

In case of this test - the assertions **and** JSON mapping take enough time, that, at least in theory, we shouldn't need to wait. Yet, relying on "JSON serialization being slow" as a determiner of test success is super flaky. Hence, the `waitFor` method.

ParaskP7: > But it's still case of timing...

Thanks for sharing a new diff, and I see, thanks for explaining again! ✅ 

I don't want to block this PR, feel free to merge it when you think it is ready, I just wanted to question it a bit just to get a better idea on the actual value of this functional test and how you would that be of use later on. 👍

parsely/build.gradle
ParaskP7: See e6727a0bd213dd72cad11c45988142894c237d42 and thus resolving this, thanks!

parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
wzieba: Thanks for the questions! We can always iterate over how those tests look now. For now, I think this test brings value - we're sure that saving and getting data from the local file is working fine and doesn't produce duplicates.

ParaskP7: 👍 💯 🥇

PR #86, opened 2023-10-31
Refactor `FlushManager` to Kotlin Coroutines

wzieba:
## Description

This PR moves `FlushTimer` class to separate class, rewrites it in Kotlin Coroutines and adds unit tests coverage. Additionally, it adds a functional test that asserts the correct flushing queue logic.

## To test
Green light from CI should be fine 👍


parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
```diff
@@ -70,6 +71,47 @@ class FunctionalTests {
...
+
+            Thread.sleep((flushInterval / 2).inWholeMilliseconds)
+
+            scenario.onActivity {
+                parselyTracker.trackPageview("url", null, null, null)
+            }
+
+            Thread.sleep((flushInterval / 2).inWholeMilliseconds)
+
+            val firstRequestPayload = server.takeRequest(2000, TimeUnit.MILLISECONDS)?.toMap()
```
wzieba, line 101: Such a long waiting time is needed because of poor performance on GitHub Actions emulator. Also, the tests are executed differently, comparing to local execution (both device and same-settings emulator) and Firebase Test Lab. You can compare the logs, that just after `[Parsely] Sending request with 2 events` log, GitHub Actions emulator sends test activity to `Paused` (`Lifecycle status change: com.parsely.parselyandroid.FunctionalTests$SampleActivity@316b36c in: PAUSED`), when it doesn't happen anywhere else...

All of this makes me think about pivoting to:
- using FIrebase Test Lab instead CI emulator
- moving from emulator-based functional test to Robolectric

parsely/build.gradle
```diff
@@ -63,12 +64,14 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.3'
     implementation 'com.google.android.gms:play-services-ads-identifier:18.0.1'
     implementation 'androidx.lifecycle:lifecycle-process:2.6.2'
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutinesVersion"
```
ParaskP7, line 67: Praise (🎉): Woohoo! 🥇

parsely/src/main/java/com/parsely/parselyandroid/FlushManager.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+internal class FlushManager(
+    private val parselyTracker: ParselyTracker,
+    val intervalMillis: Long,
+    private val coroutineScope: CoroutineScope
+) {
+    private var job: Job? = null
+
+    fun start() {
+        if (job?.isActive == true) {
```
ParaskP7, line 24: Minor (🔍): Consider making this a one-liner, hoping to further increase readability. 😊 

```kotlin
if (job?.isActive == true) return
```

PS: Same applies to:

```kotlin
    fun stop() = job?.cancel()

    fun isRunning() = job?.isActive == true
```

parsely/src/main/java/com/parsely/parselyandroid/FlushManager.kt
```diff
@@ -0,0 +1,41 @@
...
+            return
+        }
+        job = coroutineScope.launch {
```
ParaskP7, line 27: Question (❓): How confident are we that the `Timer` related implementation and the `Coroutines` related one are equivalent, from 1 to 10, like 10? 😊

```kotlin
        runningTask = object : TimerTask() {
            override fun run() {
                parselyTracker.flushEvents()
            }
        }
        parentTimer.scheduleAtFixedRate(runningTask, intervalMillis, intervalMillis)
```

vs

```kotlin
        job = coroutineScope.launch {
            while (isActive) {
                delay(intervalMillis)
                parselyTracker.flushEvents()
            }
        }
```

parsely/src/main/java/com/parsely/parselyandroid/ParselyCoroutineScope.kt
```diff
@@ -0,0 +1,7 @@
+package com.parsely.parselyandroid
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+
+val sdkScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```
ParaskP7, line 7: Question (❓): Do we actually need the `SupervisorJob` here, or is it just a nice to have? 🤔

parsely/src/androidTest/java/com/parsely/parselyandroid/FunctionalTests.kt
ParaskP7: > using FIrebase Test Lab instead CI emulator

Hmmm, maybe, not sure? 🤔 

> moving from emulator-based functional test to Robolectric

Hmmm, if I could vote, I would say to NOT making that move to Robolectric. 😊

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, great job, one more test suite live! 🌟 🧪 🌟 

-----

I have left few questions (❓) and a minor (🔍) comment for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

-----

`EXTRAS`

Suggestion (💡): For this PR, and although it seems safe to merge by just depending on CI, I would still recommend having some actual manual testing instructions, to verify them and only then to proceed with merging this, even if that's just for documentation purposes. I am recommending that because `FlushManager` has been completely re-written to Kotlin + Coroutines, plus, the test suite for it was added after the fact, not before. Just for us to be on the safe side, wdyt? 🤷

parsely/src/main/java/com/parsely/parselyandroid/FlushManager.kt
wzieba: Hm, I'm not sure how I could rate - do you have some concerns? If yes, we can add a test for concerning scenario. I think the code in Coroutines will behave exactly the same, including initial delay (second parameter of the method, [docs](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/Timer.html#scheduleAtFixedRate(java.util.TimerTask,%20long,%20long)))

ParaskP7: > Hm, I'm not sure how I could rate

From your answer it seems that you are fully confident, like a 10? 😊 

> do you have some concerns? If yes, we can add a test for concerning scenario.

No particular concerns other than me never had to do this `Timer` to `Coroutines` conversion, and me not going to deep into the implementation details of how the `Timer` works other than reading the documentation. 🤷 

> I think the code in Coroutines will behave exactly the same, including initial delay (second parameter of the method, [docs](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/Timer.html#scheduleAtFixedRate(java.util.TimerTask,%20long,%20long)))

Yeap, it does seems so indeed... 🤞 

-----

FYI: I am resolving this, no need for further conversation. 😊

parsely/src/main/java/com/parsely/parselyandroid/ParselyCoroutineScope.kt
wzieba: Children of the `SupervisorJob` can fail/be canceled independently. This `CoroutineScope` is aimed for different jobs in the SDK, e.g. `FlushTimer` shouldn't fail if HTTP request did.

ParaskP7: > Children of the SupervisorJob can fail/be canceled independently. 

💯 

> This CoroutineScope is aimed for different jobs in the SDK

Oh, didn't know about this, thanks for clarifying that for me @wzieba ! 👍

wzieba:
>I would still recommend having some actual manual testing instructions, to verify them and only then to proceed with merging this, even if that's just for documentation purposes. I am recommending that because FlushManager has been completely re-written to Kotlin + Coroutines, plus, the test suite for it was added after the fact, not before.

I agree that it's quite a big change (first of many to the SDK), but one of the reasons for adding functional tests, was to eliminate a need for manual testing. They are as close to the real scenario as possible - only instead of a person tapping buttons, we run SDK functions, and instead of a person validating logs - we have assertions. Do you have some additional scenario in mind that we should test here manually?

ParaskP7:
👋 @wzieba !

> I agree that it's quite a big change (first of many to the SDK), but one of the reasons for adding functional tests, was to eliminate a need for manual testing.

💯 but also, IMHO, elevating manual testing after merging this, not before, thus me suggested we still do manual testing on this PR. Still, I do understand were you are coming from. 💯 

> They are as close to the real scenario as possible - only instead of a person tapping buttons, we run SDK functions, and instead of a person validating logs - we have assertions.

👍 

> Do you have some additional scenario in mind that we should test here manually?

Nothing comes to mind as I am not having all testing scenarios in my head, I just did some basic manual testing myself and wondered if I covered everything needed. Thus me trying to be on the safe side and asking you for an explicit testing scenario list so that I can be a bit more relaxed with this merge. 😊

ParaskP7:
> @ParaskP7 heads up - in case of some changes/bugfixes that can be needed for current main, I've decided to change target of this PR to a new, long-running feature branch coroutines.

Oh nice, this kind of makes sense, thanks for the heads-up on that @wzieba ! 👍 

> As soon as all code will be migrated to use Coroutines, I'll merge the long-running branch to main and prepare a version of SDK.

👍

PR #88, opened 2023-11-06
Move `ParselyAPIConnection` to Kotlin

wzieba:
### Description

This small PR moves `ParselyAPIConnection` to Kotlin. I need it to be in Kotlin to introduce migrate from `AsyncTask` used in this class to Kotlin Coroutines. So please note, that this class will be heavily rewritten in next PRs.

### How to test

As this change changes code in HTTP communication, which is mocked in our tests, please:
- Update `MainActivity.java#31` and make the sample app not in the debug mode
- Tap on "track url"
- Move app to the background
- See in logs that POST request was successful


codecov.yml
```diff
@@ -1,2 +1,7 @@
 github_checks:
     annotations: false
+
+coverage:
```
ParaskP7, line 4: > We don't need such check in this SDK. Also, it can be false positive when moving from Java to Kotlin

Praise (❤️): I agree and thanks for this change!

parsely/src/main/java/com/parsely/parselyandroid/ParselyAPIConnection.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+package com.parsely.parselyandroid
+
+import android.os.AsyncTask
```
ParaskP7, line 18: Minor (🔍): Consider suppressing this deprecation as well; needs to be done on the file level (above the package declaration):

`@file:Suppress("DEPRECATION")`

parsely/src/main/java/com/parsely/parselyandroid/ParselyAPIConnection.kt
```diff
@@ -0,0 +1,62 @@
...
+import java.net.HttpURLConnection
+import java.net.URL
+
+@Suppress("DEPRECATION")
+internal class ParselyAPIConnection(private val tracker: ParselyTracker) : AsyncTask<String?, Exception?, Void?>() {
+    private var exception: Exception? = null
+
+    @Deprecated("Deprecated in Java")
```
ParaskP7, line 26: Question (❓): Why did you chose `@Deprecated(...)`, which has a different usage, instead of just using `@Suppress("OVERRIDE_DEPRECATION")`? 🤔

PS: Same applies to `onPostExecute(...)` function further below.

parsely/src/main/java/com/parsely/parselyandroid/ParselyAPIConnection.kt
```diff
@@ -0,0 +1,62 @@
...
+    override fun doInBackground(vararg data: String?): Void? {
+        var connection: HttpURLConnection? = null
```
ParaskP7, line 28: Minor (🔍): Consider updating this local variable to a non-null `val`?

`val connection: HttpURLConnection`

PS: This local variable can also be moved within the `try/catch` block it seems, which IMHO is better to have it as close to usage, right? 🤔

ParaskP7 (approved):
👋 @wzieba !

-----

I have left a questions (❓) and some minor (🔍) comments for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

parsely/src/main/java/com/parsely/parselyandroid/ParselyAPIConnection.kt
wzieba: Good point - I could make it more readable but at the same point it feels okay for now, and very soon it'll be rewritten (coroutines) so I'll address your comment there then 👍

PR #91, opened 2023-11-08
Coroutines: add event flow

wzieba:
### Description

This PR introduces Kotlin Coroutines for "add an event" flow. It covers
- adding an event to the in-memory buffer
- storing event in local storage (file)

This PR also:
- removes logic of `QueueManager` as discussed internally p1699018266041689-slack-C0533SEJ82H
- makes save to file operation thread-safe of the main thread

### How to test
1. Run the test app
2. Tap on "Track URL" multiple times. Count how many times you tapped.
3. Observe logs: you'll notice a new log `Persisting n events`, where `n` is the number of events added in 1-second time frame
4. Move app to the background
5. Assert that number `m` of events in ` Sending request with m events` log equals the number you counted in step 2


parsely/src/main/java/com/parsely/parselyandroid/InMemoryBuffer.kt
```diff
@@ -0,0 +1,43 @@
...
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+internal class InMemoryBuffer(
+    private val coroutineScope: CoroutineScope,
+    private val localStorageRepository: LocalStorageRepository,
+    private val onEventAddedListener: () -> Unit,
+) {
+
+    private val mutex = Mutex()
```
wzieba, line 17: To test if this mutual exclusion works correctly, comment this, usages in `init` and `add` and then run `stressTest`. It should fail with unexpected result - sometimes recorded events are smaller, sometimes bigger.

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
```diff
@@ -33,6 +40,10 @@ internal open class LocalStorageRepository(private val context: Context) {
         persistObject(ArrayList<Map<String, Any>>())
     }
 
+    fun remove(toRemove: List<Map<String, Any>>) {
+        persistObject(getStoredQueue() - toRemove.toSet())
+    }
```
wzieba, line 45: This method is not used in this PR. It'll be in next PRs, I just left it as I was pivoting from different ideas for scope of this PR.

.github/workflows/readme.yml
```diff
@@ -72,4 +72,6 @@ jobs:
         if: always()
         with:
           name: artifact
-          path: ./parsely/build/reports/*
+          path: |
+            ./parsely/build/reports/*
+            ./parsely/build/outputs/androidTest-results
```
ParaskP7, line 77: Praise (❤️): Nice!

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
```diff
@@ -21,6 +27,7 @@ internal open class LocalStorageRepository(private val context: Context) {
             val oos = ObjectOutputStream(fos)
             oos.writeObject(o)
             oos.close()
+            fos.close()
```
ParaskP7, line 30: Praise (❤️): Good catch, kudos! 🥇 

PS: Same apply to `getStoredQueue()`.

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
ParaskP7: Good to know, thanks for clarifying that for me @wzieba ! 🙇

A follow-up question I had was related to the `purgeStoredQueue()` related test that got (kind of) removed, as it was placed with the `remove(...)` test. Will `purgeStoredQueue()` be deleting in next PRs maybe, if not I suggest keeping that test? 🤔

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
```diff
@@ -70,10 +82,9 @@ internal open class LocalStorageRepository(private val context: Context) {
     /**
      * Save the event queue to persistent storage.
      */
-    @Synchronized
-    open fun persistQueue(inMemoryQueue: List<Map<String, Any?>?>) {
-        ParselyTracker.PLog("Persisting event queue")
-        persistObject((inMemoryQueue + getStoredQueue()).distinct())
+    open suspend fun insertEvents(toInsert: List<Map<String, Any?>?>) = mutex.withLock {
```
ParaskP7, line 85: Question (❓): Is the `mutex` usage here and on this repository really needed? 🤔

1. I am asking this because this `insertEvents(...)` is having only one usage, it being called from within the `InMemoryBuffer.init { ... }` block, which is already guarding by another `mutex`. 🤔
2. If that `mutex` is really needed, just because you want to make the repository thread safe (04877191351c14cd440837652fd2eacb0d5d6905), following the same logic, you might want to do the same for the `enqueueEvent(...)` and lock that with a `mutex` as well, no? 🤔
3. IMHO having 2 `mutex` might end-up complicating things a bit, especially if they are not really needed, at least for now. Having said that, I am not sure what are your plans with this repository. Thus, please take this with a gain of salt, I just wanted to raise this question now, even if it is a dump one, instead of ignoring this whole thing altogether. 🙏

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
```diff
@@ -70,10 +82,9 @@ internal open class LocalStorageRepository(private val context: Context) {
...
+        ParselyTracker.PLog("Persisting ${toInsert.size} events")
```
ParaskP7, line 86: Suggestion (💡): Just like it is done on `InMemoryBuffer.add(...)`, maybe this log can be moved to `InMemoryBuffer.init { ... }` as well and make the logging closer to source and more consistent, at least with each other. Wdyt? 🤔

parsely/src/main/java/com/parsely/parselyandroid/ParselyCoroutineScope.kt
```diff
@@ -1,7 +1,9 @@
 package com.parsely.parselyandroid
 
+import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 
-val sdkScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
+internal val sdkScope =
+    CoroutineScope(SupervisorJob() + Dispatchers.IO + CoroutineName("Parse.ly SDK Scope"))
```
ParaskP7, line 9: Praise (❤️): Nicely done of the Coroutine name, I like! 🥇

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
```diff
@@ -75,16 +76,20 @@ protected ParselyTracker(String siteId, int flushInterval, Context c) {
         context = c.getApplicationContext();
         eventsBuilder = new EventsBuilder(context, siteId);
         localStorageRepository = new LocalStorageRepository(context);
+        flushManager = new FlushManager(this, flushInterval * 1000L,
+                ParselyCoroutineScopeKt.getSdkScope());
+        inMemoryBuffer = new InMemoryBuffer(ParselyCoroutineScopeKt.getSdkScope(), localStorageRepository, () -> {
+            if (!flushTimerIsActive()) {
```
ParaskP7, line 82: Praise (❤️): Love the listener addition, make the whole implementation much more simpler and at the same time harder to mess with, kudos! 🥇

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.java
```diff
@@ -553,24 +546,19 @@ private class FlushQueue extends AsyncTask<Void, Void, Void> {
...
             }
-            HashSet<Map<String, Object>> hs = new HashSet<>();
-            ArrayList<Map<String, Object>> newQueue = new ArrayList<>();
 
-            hs.addAll(eventQueue);
-            hs.addAll(storedQueue);
-            newQueue.addAll(hs);
             PLog("Flushing queue");
-            sendBatchRequest(newQueue);
+            sendBatchRequest(storedQueue);
```
ParaskP7, line 561: Praise (❤️): Removing `eventQueue` makes everything so much simpler! 💯

parsely/src/test/java/com/parsely/parselyandroid/InMemoryBufferTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(RobolectricTestRunner::class)
+internal class InMemoryBufferTest {
+
+    private lateinit var sut: InMemoryBuffer
+    private val repository = FakeLocalStorageRepository()
+
+    @Test
+    fun `when adding a new event, then save it to local storage and run onEventAdded listener`() = runTest {
```
ParaskP7, line 23: Suggestion (💡): Consider splitting this test into two:

1. `when adding a new event, then save it to local storage`
2. `given a listener, when adding a new event, then run listener`

parsely/src/main/java/com/parsely/parselyandroid/InMemoryBuffer.kt
ParaskP7: Thanks for the instructions! 🥇 

FYI: I run that `stressTest` 3 times after removing `mutex` and everything was still working as expected on my side (testing on actual device). I then also triggered the whole `./gradlew connectedDebugAndroidTest` from CLI and it still worked as expected, everything was successful. 🤷

ParaskP7: UPDATE: I then also removed the `mutex` on the repository level and run the `stressTest` 3 times, along with running the whole `./gradlew connectedDebugAndroidTest` from CLI. Running the `stressTest` succeeded every time, but running the whole `./gradlew connectedDebugAndroidTest` from CLI indeed failed (both times I tried):

```
com.parsely.parselyandroid.FunctionalTests > stressTest[Pixel 4 - 13] FAILED 
        java.lang.AssertionError:
        Expected size: 500 but was: 1000 in:
Tests on Pixel 4 - 13 failed: There was 1 failure(s).
```

PS: 
1. I then uncommented the `mutex` on the repository level and run whole `./gradlew connectedDebugAndroidTest` from CLI again and it started failing as well. 👍 
2. I then, uncommented the `mutex` on the buffer level and  run whole `./gradlew connectedDebugAndroidTest` from CLI again and it now succeeded. ✅ 

I guess there exist some flakiness but it seems that everything is working as expected after all. Just wanted to document my testing here, just in case it somehow helps you.

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, awesome job with `InMemoryBuffer` and removing the `eventQueue`, lots of complexity is now actually gone! 🌟 🎉 🚀 

-----

I have left a questions (❓) and a couple of suggestions (💡) for you to consider. I am going to approve this PR anyway, since none is blocking (ish). I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
wzieba: Good question! I should've left more context here. At this very moment, removing this `mutex` shouldn't change much - the only usage of this method is covered by `mutex` in `InMemoryQueue` already, as you rightly notice (point 1).

But, with next changes, I'm going to make the whole `LocalStorageRepository` thread safe (point 2) and this `mutex` will be needed - it'll make sure that events are not e.g. inserted at the same time when we remove other events. I couldn't add this right away as using `mutex` makes methods `suspend`, making the scope of the PR too big. How does it sound to you?

ParaskP7: > How does it sound to you?

Sound good @wzieba , it seems that everything is under control and that you have a solid plan going forward, a 👍 from me, thanks for the additional context, much appreciated. 🙇

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
wzieba: > Will purgeStoredQueue() be deleting in next PRs maybe,

Yes, exactly - I'd prefer to have a granular control of what's removed from the local storage.

parsely/src/main/java/com/parsely/parselyandroid/InMemoryBuffer.kt
wzieba: Thanks for sharing! I think for some reason, they're more likely to fail without `mutex` on emulator. In any case, one can also remove the `delay` to the `InMemoryBuffer#init` - this would make the test fail more certain.

ParaskP7: > I think for some reason, they're more likely to fail without mutex on emulator. 

Yea, maybe! 🤷 

> In any case, one can also remove the delay to the InMemoryBuffer#init - this would make the test fail more certain.

True, good tip! 💯

PR #92, opened 2023-11-11
Coroutines: send events flow

wzieba:
TBD

### DIFF
--- a/parsely/src/main/java/com/parsely/parselyandroid/FlushManager.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/FlushManager.kt
@@ -13,14 +13,14 @@ import kotlinx.coroutines.launch
  * Handles stopping and starting the flush timer. The flush timer
  * controls how often we send events to Parse.ly servers.
  */
-internal class FlushManager(
+internal open class FlushManager(
     private val parselyTracker: ParselyTracker,
     val intervalMillis: Long,
     private val coroutineScope: CoroutineScope
 ) {
     private var job: Job? = null
 
-    fun start() {
+    open fun start() {
         if (job?.isActive == true) return
 
         job = coroutineScope.launch {
@@ -31,8 +31,8 @@ internal class FlushManager(
         }
     }
 
-    fun stop() = job?.cancel()
+    open fun stop() = job?.cancel()
 
-    val isRunning: Boolean
+    open val isRunning: Boolean
         get() = job?.isActive ?: false
 }
--- a/parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
@@ -32,7 +32,7 @@ internal open class LocalStorageRepository(private val context: Context) {
         }
     }
 
-    suspend fun remove(toRemove: List<Map<String, Any?>?>) = mutex.withLock {
+    open suspend fun remove(toRemove: List<Map<String, Any?>?>) = mutex.withLock {
         persistObject(getStoredQueue() - toRemove.toSet())
     }
 
--- a/parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
@@ -5,7 +5,6 @@ import java.io.EOFException
 import java.io.FileNotFoundException
 import java.io.ObjectInputStream
 import java.io.ObjectOutputStream
-import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 
@@ -31,13 +30,6 @@ internal open class LocalStorageRepository(private val context: Context) {
         } catch (ex: Exception) {
             ParselyTracker.PLog("Exception thrown during queue serialization: %s", ex.toString())
         }
-    }
-
-    /**
-     * Delete the stored queue from persistent storage.
-     */
-    fun purgeStoredQueue() {
-        persistObject(ArrayList<Map<String, Any>>())
     }
 
     suspend fun remove(toRemove: List<Map<String, Any?>?>) = mutex.withLock {
@@ -69,14 +61,6 @@ internal open class LocalStorageRepository(private val context: Context) {
             )
         }
         return storedQueue
-    }
-
-    /**
-     * Delete an event from the stored queue.
-     */
-    open fun expelStoredEvent() {
-        val storedQueue = getStoredQueue()
-        storedQueue.removeAt(0)
     }
 
     /**
--- a/parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
@@ -40,7 +40,7 @@ internal open class LocalStorageRepository(private val context: Context) {
         persistObject(ArrayList<Map<String, Any>>())
     }
 
-    fun remove(toRemove: List<Map<String, Any>>) {
+    suspend fun remove(toRemove: List<Map<String, Any?>?>) = mutex.withLock {
         persistObject(getStoredQueue() - toRemove.toSet())
     }
 
--- a/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
@@ -35,7 +35,7 @@ internal class SendEvents(
                     .fold(
                         onSuccess = {
                             ParselyTracker.PLog("Pixel request success")
-                            parselyTracker.purgeEventsQueue()
+                            localStorageRepository.purgeStoredQueue()
                             ParselyTracker.PLog("Event queue empty, flush timer cleared.")
                             parselyTracker.stopFlushTimer()
                         },
--- a/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
@@ -1,34 +1,38 @@
 package com.parsely.parselyandroid
 
 import com.parsely.parselyandroid.JsonSerializer.toJson
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 
 internal class SendEvents(
     private val parselyTracker: ParselyTracker,
-    private val localStorageRepository: LocalStorageRepository
+    private val localStorageRepository: LocalStorageRepository,
+    private val parselyAPIConnection: ParselyAPIConnection,
+    private val scope: CoroutineScope
 ) {
 
     operator fun invoke(isDebug: Boolean) {
-        val eventsToSend = localStorageRepository.getStoredQueue()
+        scope.launch {
+            val eventsToSend = localStorageRepository.getStoredQueue()
 
-        if (eventsToSend.isEmpty()) {
-            return
+            if (eventsToSend.isEmpty()) {
+                return@launch
+            }
+            ParselyTracker.PLog("Sending request with %d events", eventsToSend.size)
+
+            val batchMap: MutableMap<String, Any> = HashMap()
+            batchMap["events"] = eventsToSend
+            val jsonPayload = toJson(batchMap).orEmpty()
+
+            if (isDebug) {
+                ParselyTracker.PLog("Debug mode on. Not sending to Parse.ly")
+                localStorageRepository.purgeStoredQueue()
+            } else {
+                ParselyTracker.PLog("Requested %s", ParselyTracker.ROOT_URL)
+
+                parselyAPIConnection.send(jsonPayload)
+            }
+            ParselyTracker.PLog("POST Data %s", toJson(batchMap))
         }
-        ParselyTracker.PLog("Sending request with %d events", eventsToSend.size)
-
-        val batchMap: MutableMap<String, Any> = HashMap()
-        batchMap["events"] = eventsToSend
-        val jsonPayload = toJson(batchMap)
-
-        if (isDebug) {
-            ParselyTracker.PLog("Debug mode on. Not sending to Parse.ly")
-            localStorageRepository.purgeStoredQueue()
-        } else {
-            ParselyTracker.PLog("Requested %s", ParselyTracker.ROOT_URL)
-            ParselyAPIConnection(parselyTracker).execute(
-                ParselyTracker.ROOT_URL + "mobileproxy",
-                jsonPayload
-            )
-        }
-        ParselyTracker.PLog("POST Data %s", toJson(batchMap))
     }
 }
--- a/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
@@ -1,6 +1,6 @@
 package com.parsely.parselyandroid
 
-import com.parsely.parselyandroid.JsonSerializer.toJson
+import com.parsely.parselyandroid.JsonSerializer.toParselyEventsPayload
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
@@ -20,9 +20,7 @@ internal class SendEvents(
             }
             ParselyTracker.PLog("Sending request with %d events", eventsToSend.size)
 
-            val batchMap: MutableMap<String, Any> = HashMap()
-            batchMap["events"] = eventsToSend
-            val jsonPayload = toJson(batchMap).orEmpty()
+            val jsonPayload = toParselyEventsPayload(eventsToSend)
 
             ParselyTracker.PLog("POST Data %s", jsonPayload)
 
--- a/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
@@ -5,7 +5,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
 internal class SendEvents(
-    private val parselyTracker: ParselyTracker,
+    private val flushManager: FlushManager,
     private val localStorageRepository: LocalStorageRepository,
     private val parselyAPIConnection: ParselyAPIConnection,
     private val scope: CoroutineScope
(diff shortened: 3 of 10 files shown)


parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
```diff
@@ -33,23 +32,21 @@ internal open class LocalStorageRepository(private val context: Context) {
         }
     }
 
-    /**
-     * Delete the stored queue from persistent storage.
-     */
-    fun purgeStoredQueue() {
-        persistObject(ArrayList<Map<String, Any>>())
-    }
+    open suspend fun remove(toRemove: List<Map<String, Any?>?>) {
```
oguzkocer, line 35: _This comment is not about this specific line, but I think this is the best place available in this PR review._

There is a slight inconsistency about where the Mutex lock is applied within this class. It seems every function has the lock applied, except for `persistObject`. However, `persistObject` is also always called from within a lock.

Also, it seems the mutex lock is sometimes called from an already locked state as such:
```
mutex.withLock {
  mutex.withLock {
  }
}
```
I am not entirely sure how this is handled internally, but at the very least, it seems a bit inefficient. This inefficiency might be necessary in order to keep things safe, I am not sure.

I think it'd be good to look into how Mutex lock is used as a whole to make it more consistent and possibly more efficient. But, maybe it could be done in its own PR?

parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
```diff
@@ -0,0 +1,57 @@
...
+internal class SendEvents(
+    private val flushManager: FlushManager,
+    private val localStorageRepository: LocalStorageRepository,
+    private val parselyAPIConnection: ParselyAPIConnection,
+    private val scope: CoroutineScope
+) {
+
+    private val mutex = Mutex()
+
+    operator fun invoke(isDebug: Boolean) {
```
oguzkocer, line 18: How do you feel about renaming `isDebug` to `skipSendingEvents` or something similar?

`isDebug` could mean anything and when such an argument is available, sometimes it's used for several different things within the implementation. Having a specific name should help with self-documentation and it'll discourage developers from overloading the argument with multiple responsibilities.

In this context, I think it's kind of clear what `isDebug` is used for - but I think a specific name will be more future proof.

parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
```diff
@@ -0,0 +1,57 @@
...
+
+                if (eventsToSend.isEmpty()) {
+                    flushManager.stop()
+                    return@launch
+                }
+                ParselyTracker.PLog("Sending request with %d events", eventsToSend.size)
+
+                val jsonPayload = toParselyEventsPayload(eventsToSend)
+
+                ParselyTracker.PLog("POST Data %s", jsonPayload)
```
oguzkocer, line 31: Nitpick: These can probably go into the `else` clause - although if you do that you might want to add the number of events to the log in `isDebug` clause.

A further improvement might be to return from `isDebug` which should help with the readability. It'd also be more consistent with the prior `if (eventsToSend.isEmpty())` check.

parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
```diff
@@ -0,0 +1,57 @@
...
+                    ParselyTracker.PLog("Requested %s", ParselyTracker.ROOT_URL)
+                    parselyAPIConnection.send(jsonPayload)
+                        .fold(
+                            onSuccess = {
+                                ParselyTracker.PLog("Pixel request success")
+                                localStorageRepository.remove(eventsToSend)
+                                ParselyTracker.PLog("Event queue empty, flush timer cleared.")
+                                if (localStorageRepository.getStoredQueue().isEmpty()) {
+                                    flushManager.stop()
+                                }
```
oguzkocer, line 46: Would it make sense to skip this part and let the `if (eventsToSend.isEmpty())` check handle this logic for us?

oguzkocer (commented):
@wzieba I've completed reviewing the implementation part of the PR. I haven't reviewed the tests yet or tested the implementation because if you decide the address any of the comments, it'd be best to review the tests with those included.

I am hoping that I am not completely off with my questions and suggestions, but if I am, please excuse me as I am getting back to product development PR reviews 😰

parsely/src/main/java/com/parsely/parselyandroid/SdkInit.kt
```diff
@@ -0,0 +1,18 @@
+package com.parsely.parselyandroid
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+internal class SdkInit(
+    private val scope: CoroutineScope,
+    private val localStorageRepository: LocalStorageRepository,
+    private val flushManager: FlushManager,
```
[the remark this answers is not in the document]
wzieba: Sure, good idea - there's less boilerplate with the simple function 👍 Addressed in 366dae5669de3be03dbb73dd4b84180aabe1cf87

parsely/src/main/java/com/parsely/parselyandroid/FlushManager.kt
```diff
@@ -13,14 +13,14 @@ import kotlinx.coroutines.launch
  * Handles stopping and starting the flush timer. The flush timer
  * controls how often we send events to Parse.ly servers.
  */
-internal class FlushManager(
+internal open class FlushManager(
```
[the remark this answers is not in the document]
wzieba: I'm fine with introducing an interface - my concern for not doing that, was that it requires a little more code. After rethinking it with your comment, I can see a nice advantage though, that when creating a fake object, we won't have to call the constructor of the real object (and provide possibly irrelevant information for a test).

Addressed in:
- acae5f6
- 77303c6
- be74b28
- a13485c

parsely/src/main/java/com/parsely/parselyandroid/SdkInit.kt
```diff
@@ -0,0 +1,18 @@
...
+) {
+    fun initialize() {
+        scope.launch {
+            if (localStorageRepository.getStoredQueue().isNotEmpty()) {
```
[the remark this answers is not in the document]
wzieba: > I think calling flushManager.start() might be simpler and more efficient.

Yes, definitely - nicely spotted! I, too, think that `flushManager.start()` without additional check will be just fine. Addresed in cb5113c

parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
wzieba: I agree with your arguments: +1 from me for renaiming `isDebug` to `skipSendingEvents`. Additionally, I renamed `SendEvents` to `FlushQueue` because I believe the possibility to run "send events with skipping sending events" might be confusing. Addressed in c9872d4.

parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
wzieba: Sure, good point 👍 addressed in 269a9d6 and 409c42d

parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
wzieba: I'm not sure: between `eventsToSend.isEmpty` check and sending events, the consumer app might report new events, which are saved to the stored queue. In such case, we don't want to stop the flush timer. That's why I added check for local storage just before deciding if we want to stop flush manager. 

Does it make sense, WDYT?

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
wzieba: Thanks for this comment. I'm totally fine with changing this logic, but I'd like to explain my rationale on those decisions.

> It seems every function has the lock applied, except for `persistObject`.

True, and it's intended. As you rightfully notice, `mutex` of `LocalStorageRepository` locks execution of all public methods. As we control all possible executions of `private persistObject`, it's safe to leave it without `mutex`. Also, adding the same lock to this method would create a thread lock ([sample](https://pl.kotl.in/p9DbWDKUb)).

> Also, it seems the mutex lock is sometimes called from an already locked state as such:

It's crucial to emphasize, that in the SDK those two `mutex`es are different objects and have different purpose. If it was the same object, we would have a thread lock (sample above). Adding more context to the example you provided:

```
mutexA.withLock { (1)
  mutexB.withLock { (2)
        persistObject
  }
  //continue with execution locked by `mutexA` only
}
```

`mutexA` is from `SendEvents`/`FlushQueue` use case. It doesn't allow running this use case twice at the same time. Otherwise, if we run it twice, we're at risk of sending duplicated events (first execution wouldn't call `remove` method before second execution called `getStoredQueue`). I could address it in implementation of the use case by e.g. introducing `LocalStorageRepository#getAndRemoveStoredQueue` but I thought `Mutex` is also a fine solution.

Also, you might want to take a look at `InMemoryBuffer` - in case of this class, the `mutexA` is there, but the reasoning is the same (although I don't have an idea for a different, no-mutex approach there)

`mutexB` is from `LocalStorageRepository`. It doesn't allow manipulating the local file at the same time by two different methods. It's crucial to not risk corrupting the local file and safely performing operations.

> I think it'd be good to look into how Mutex lock is used as a whole to make it more consistent and possibly more efficient. But, maybe it could be done in its own PR?

Sure - we could probably remove `mutexA` (I mean: the one in `SendEvents`/`FlushQueue` class) but I'm not sure if it's crucial for performance. If you see other improvements, I'm eager to discuss them! If you'd like to experiment with `Mutex` usage, I can recommend running `FunctionalTests#stressTest` for verification if coroutines are not locking and/or are efficient.

wzieba:
Thank you @oguzkocer for the review! I've applied some of the comments and replied to other comments so we can continue discussion. Ready for re-review.

parsely/src/main/java/com/parsely/parselyandroid/FlushQueue.kt
```diff
@@ -0,0 +1,56 @@
...
+
+                if (eventsToSend.isEmpty()) {
+                    flushManager.stop()
+                    return@launch
+                }
+
+                if (skipSendingEvents) {
+                    ParselyTracker.PLog("Debug mode on. Not sending to Parse.ly. Otherwise, would sent ${eventsToSend.size} events")
+                    repository.remove(eventsToSend)
+                    return@launch
```
oguzkocer, line 31: Does this need to stop the `flushManager` as well? I guess it'll get stopped the next time it's flushed. 🤔 

I wonder if we should skip starting the `FlushQueue` all together in a case where we don't want the events to be sent. This is probably beyond the scope of this PR though, so I am just making a note of it here.

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
oguzkocer: > It's crucial to emphasize, that in the SDK those two `mutex`es are different objects and have different purpose. If it was the same object, we would have a thread lock (sample above). Adding more context to the example you provided:

Yeap, I was aware of that, and I should have specified that in my pseudocode as I can see how it could be interpreted as the same locks. Sorry about that!

> True, and it's intended. As you rightfully notice, `mutex` of `LocalStorageRepository` locks execution of all public methods. As we control all possible executions of `private persistObject`, it's safe to leave it without `mutex`.

I think in the context of "some of these functions are meant to be publicly consumed", trade-off makes sense to me. I was thinking in the lines of having an internal only storage that is exposed to the public through an encapsulating type. However, I didn't mention that at all and I am not even sure there is enough benefit to the indirection.

---

I think the case I should have brought up as concrete example is the mutex usage in `insertEvents` & `getStoredQueue`. In this case, we are locking the mutex, reading the disk, releasing the lock, then locking it immediately after to persist objects and then finally releasing it again. This type of relationship may be necessary in some cases when part of the relationship is outside of our control. However, in this case, we are able to control this, so I think it'd be good to lock the mutex once, read the events, persist the objects, and then release the lock. Here is what I am thinking:

```
    private fun getInternalStoredQueue(): ArrayList<Map<String, Any?>?> {
        var storedQueue: ArrayList<Map<String, Any?>?> = ArrayList()
        try {
            val fis = context.applicationContext.openFileInput(STORAGE_KEY)
            val ois = ObjectInputStream(fis)
            @Suppress("UNCHECKED_CAST")
            storedQueue = ois.readObject() as ArrayList<Map<String, Any?>?>
            ois.close()
            fis.close()
        } catch (ex: EOFException) {
            // Nothing to do here.
        } catch (ex: FileNotFoundException) {
            // Nothing to do here. Means there was no saved queue.
        } catch (ex: Exception) {
            ParselyTracker.PLog(
                "Exception thrown during queue deserialization: %s",
                ex.toString()
            )
        }
        return storedQueue
    }

    override suspend fun getStoredQueue(): ArrayList<Map<String, Any?>?> = mutex.withLock {
        return getInternalStoredQueue()
    }

```

To connect the dots, this setup is actually pretty similar to what I mentioned related to "internal only storage" above. Instead of having a second type to separate the internal functions, we are utilizing private functions. Although, If we did have an internal only type that doesn't have a mutex, we'd have been able to use that one in other parts of our internal implementation and avoid the mutex inside another mutex type of situation. It's also generally a design that I like, but this is definitely outside of this PR and I am just sharing my general thoughts on the subject.

Looking forward to hear your thoughts!

parsely/src/test/java/com/parsely/parselyandroid/FlushManagerTest.kt
```diff
@@ -1,49 +1,48 @@
...
-import org.junit.runner.RunWith
-import org.robolectric.RobolectricTestRunner
 
 @OptIn(ExperimentalCoroutinesApi::class)
-@RunWith(RobolectricTestRunner::class)
 class FlushManagerTest {
 
     private lateinit var sut: FlushManager
-    private val tracker = FakeTracker()
+    private var flushEventsCounter = 0
```
oguzkocer, line 15: Do we need these properties, or would it be possible to use local variables instead? I skimmed over the tests and I can't see the need for the properties, but I might be missing something.

parsely/src/test/java/com/parsely/parselyandroid/FlushQueueTest.kt
```diff
@@ -0,0 +1,247 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(RobolectricTestRunner::class)
+class FlushQueueTest {
+
+    private lateinit var sut: FlushQueue
```
oguzkocer, line 15: Similar to my previous comment, do we need to use a property or could we use a local variable instead?

oguzkocer (commented):
@wzieba I've reviewed the tests and the example app changes and left a minor comment.

I initially had a few more nitpicks about using named parameters for coroutine scope in tests, adding an extra validation for events are sent in `FlushQueueTest` etc, but I don't think those are adding enough value for us to spend extra time in the PR review process, so I decided to remove them.

We have a couple open discussions, so I haven't tested the changes yet. Once everything is finalized, I'll test it and approve the PR. I hope the review so far has been helpful.

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
wzieba: > I think it'd be good to lock the mutex once, read the events, persist the objects, and then release the lock

True - I see your point. I've updated the code in c32302f to lock the mutex once - it seems even safer this way, as e.g. in `insertEvent`, between `getStoredEvents` and `mutex.withLock` a new event *might* be added to the queue in a different lock, making the final list provided to `persistObject` incorrect. Thanks!

> If we did have an internal only type that doesn't have a mutex, we'd have been able to use that one in other parts of our internal implementation and avoid the mutex inside another mutex type of situation.

`LocalStorageRepository` is internal (in a sense that it's impossible for SDK consumer to trigger any methods on it). Do you think about a different definition of "internal" here? 

Thinking about removing the necessity of mutex nesting, I don't see a clear solution with file-based storage. At a time of querying or saving data, we can't allow for other operations to be performed on the file and there's a constant possibility that e.g. `InMemoryBuffer` will try to save new events.

But I can also be tunnel-vissioned about the proposed approach. I think it'd be more clear for me if we discussed it with some code snippets, maybe in context of #94 ?

parsely/src/main/java/com/parsely/parselyandroid/SendEvents.kt
wzieba: Right! 🙂 Thanks, improves readability. Addressed in: 2c0ce6e

parsely/src/main/java/com/parsely/parselyandroid/FlushQueue.kt
wzieba: I removed stopping `flushManager` from this part of code in 784a021. 

> I wonder if we should skip starting the FlushQueue all together in a case where we don't want the events to be sent.

I see your point but the `skipSendingEvents` option is used only for tests (debug mode). As it's like `--dry-run`, I wouldn't opt to remove starting `flushManager` then, as this would cause a significant discrepancy between debug and non-debug modes of SDK. In other words: in debug mode, I believe we want to be as close as possible to non-debug. WDYT?

wzieba:
Thank you @oguzkocer for second round! I think I answered all your comments, ready for the next review.

parsely/src/main/java/com/parsely/parselyandroid/LocalStorageRepository.kt
oguzkocer: I think we have made some good improvements to the Mutex usage and I am fine with leaving it as is to keep the more important parts of the project moving.

> Thinking about removing the necessity of mutex nesting, I don't see a clear solution with file-based storage. At a time of querying or saving data, we can't allow for other operations to be performed on the file and there's a constant possibility that e.g. `InMemoryBuffer` will try to save new events.

I think in the ideal scenario, all writes for our storage would go through a single mutex. So, no 2 things could write to it at once. We'd also only let someone read the storage if there is no write access happening at the same time. This would be much closer to how a real DB works. Having said that, it's quite an involved process to get this right and I am not sure if there is enough benefit in trying to optimize it to the fullest.

If the current solution works out fine and doesn't have any major performance concerns, keeping it simpler seems like the way to go for us.

oguzkocer (approved):
@wzieba Everything looks great. I've tested the example app using the instructions from PR description and I've observed all the logs except for `Event queue empty, flush timer cleared.` which was removed in 784a0211aaddccc3ce22287cac3f8927c6463268.

:shipit:

PR #93, opened 2023-11-14
Refactor retrieving advertisement id to Coroutines

wzieba:
It's probably better to review this PR after #95 .

## Description

This PR moves logic from `GetAdKey: AsyncTask` into Coroutines-based `AdvertisementIdProvider`.

It also refactors and unifies getting another id: `ANDROID_ID`, which should be used if the device is missing Google Play Services.

## How to test

Added and improved unit tests in this PR should be just fine in validating the introduced change.


parsely/src/main/java/com/parsely/parselyandroid/AdvertisementIdProvider.kt
```diff
@@ -0,0 +1,47 @@
+package com.parsely.parselyandroid
+
+import android.content.Context
+import android.provider.Settings
+import com.google.android.gms.ads.identifier.AdvertisingIdClient
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+internal class AdvertisementIdProvider(
```
oguzkocer, line 9: I am curious about the general design choice for this class.

If I am understanding it correctly, when an object of this class is instantiated, if we were to read the `adKey` immediately, it'll return `null`. Looking at the constructor signature, there is a hint that we are using coroutines for _something_ but there isn't much context about what it's used for unless we look into the implementation. So, I think a developer may not realize that gotcha and end up requesting the `adKey` before it's ready. Also, I am not even sure if there is a straightforward way to tell when the `adKey` is ready to be retrieved.

So, I was wondering if we should use a `suspend` function for `provide` so that it can be used in a proper coroutine scope without worrying about the internal details. What do you think?

parsely/src/main/java/com/parsely/parselyandroid/EventsBuilder.java
```diff
@@ -96,90 +82,4 @@ Map<String, Object> buildEvent(
...
-        PLog("adkey is: %s, uuid is %s", adKey, getSiteUuid());
-        final String uuid = (adKey != null) ? adKey : getSiteUuid();
-        dInfo.put("parsely_site_uuid", uuid);
-        dInfo.put("manufacturer", android.os.Build.MANUFACTURER);
-        dInfo.put("os", "android");
-        dInfo.put("os_version", String.format("%d", android.os.Build.VERSION.SDK_INT));
-
-        // FIXME: Not passed in event or used anywhere else.
-        CharSequence txt = context.getPackageManager().getApplicationLabel(context.getApplicationInfo());
-        dInfo.put("appname", txt.toString());
```
oguzkocer, line 117: Was this intentionally removed? If so, that's OK - just want to make sure it's not an oversight.

parsely/src/test/java/com/parsely/parselyandroid/AndroidDeviceInfoRepositoryTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.robolectric.shadows.ShadowBuild
+
+private const val SDK_VERSION = 33
+private const val MANUFACTURER = "test manufacturer"
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [SDK_VERSION])
+internal class AndroidDeviceInfoRepositoryTest {
+
+    private lateinit var sut: AndroidDeviceInfoRepository
```
oguzkocer, line 18: Does this need to be a property or could we use a local value for it?

oguzkocer (commented):
@wzieba I've finished my first pass and left a few questions for you. Let me know what you think!

parsely/src/test/java/com/parsely/parselyandroid/AndroidDeviceInfoRepositoryTest.kt
oguzkocer: That's totally OK, thanks for the changes!

I try and avoid the `var` keyword as much as possible, so anytime I see that in a PR review, I make it a point to check if we could remove it. I do something similar with properties, but I think sometimes they are the practical choice and commenting on those can get nitpicky, so I don't do it as much. Just thought I'd provide some context into how I review these.

parsely/src/main/java/com/parsely/parselyandroid/AdvertisementIdProvider.kt
oguzkocer: I think this one is important to get right. We don't have to use a suspended function if it's going to be more trouble than it's worth - but at the very least, I think it's worth having a big disclaimer in both the class definition and where it's consumed. The reason I strongly feel this way is because if we have any issues related to this, it can be hard to debug if the developer doesn't have any prior context. To be fair - the fact that it's marked as `nullable` may give enough context, so maybe it's not as big of a deal as I am making it out to be 🤷‍♂️ 

I don't think we should make any major changes in this PR, because it'll get hard to review it. However, if I was working on this, I'd give using suspended function a chance as a separate PR. Assuming we keep the current architecture, this will cause a chain reaction and the following methods may also need to become suspended functions: `AndroidDeviceInfoRepository.collectDeviceInfo`, `EventsBuilder.buildEvent` & `ParselyTracker.trackPlay`.

* `ParselyTracker.trackPlay`: Looking at the name, I assumed that this function would eventually become a suspended function regardless of the decision we'll make here, but looking at the implementation, I am not so sure. I'd love to hear your thoughts!
* `EventsBuilder.buildEvent`: This probably shouldn't be a suspended function, but instead take the device info as a parameter. This is used in `ParselyTracker.trackPlay` & `ParselyTracker.trackPageview` both of which _could_ be suspended functions, so I think that'll work out.
* `AndroidDeviceInfoRepository.collectDeviceInfo`: Making this a suspended function makes sense to me. It communicates that collecting device info is not an immediate thing which could be for many reasons. We could be reading from file, checking a DB etc to collect this info. Admittedly some of the info is available immediately, so I could see an argument for keeping it a regular function as well. 🤷‍♂️ 

I think the decision will heavily depend on your vision for the SDK in general and I don't want to influence that vision too much. If you think it's better to leave the current design as is, I can live with that as long as we document it.

Hope you find my thoughts useful - let me know what you think!

oguzkocer (approved):

parsely/src/main/java/com/parsely/parselyandroid/AdvertisementIdProvider.kt
oguzkocer: > As that's the case - would you mind accepting this PR, or do you think we should iterate on something more here?

I just wanted this conversation to resolve before approving the PR - in case you wanted to add a comment about this issue. Approved now - thanks for the discussion!

PR #95, opened 2023-11-29
Engagement manager to coroutines

wzieba:


parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.kt
```diff
@@ -0,0 +1,89 @@
...
+    private val parselyTracker: ParselyTracker,
+    private var latestDelayMillis: Long,
+    private val baseEvent: Map<String, Any>,
+    private val intervalCalculator: HeartbeatIntervalCalculator,
+    private val coroutineScope: CoroutineScope,
+    private val clock: Clock,
+) {
+    private var job: Job? = null
+    private var totalTime: Long = 0
+    private var startTime: Duration
```
oguzkocer, line 31: I think we can use a local variable for this.

Since we are setting it at the beginning of `start()`, I don't think we need to do it in `init`. Once that's removed, we'll only be using it in `start()` and can be converted to a local variable.

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.kt
```diff
@@ -0,0 +1,89 @@
...
+                delay(latestDelayMillis)
+                doEnqueue(clock.now.inWholeMilliseconds)
+            }
+        }
+    }
+
+    fun stop() {
+        job?.let {
+            it.cancel()
+            doEnqueue(nextScheduledExecution)
```
oguzkocer, line 56: Could you provide me some context about what this line is accomplishing, so I can provide better feedback?

oguzkocer (approved):
Looks good! I've left one minor suggestion and a question which might prompt another suggestion, but I don't think any of it is blocker, so I am going to approve the PR.

For full transparency, I had some other feedback that I was working on, but then I realized the feedback is about the original implementation and not this conversion - which wouldn't be fair. In my opinion it's a good thing to keep the conversion PRs as close to the original as possible and separate any major improvements to its own PRs. So, I decided to hold off on anything that's unrelated to this specific PR.

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.kt
wzieba: Right, good point, thanks! Addressed in d282f8b1117103d251ce3dc2a213b3a27c72f3b3

wzieba:
Thank you @oguzkocer !

>  In my opinion it's a good thing to keep the conversion PRs as close to the original as possible and separate any major improvements to its own PRs.

Yes, I agree - that's exactly my intention for those conversion PRs. Thank you for thinking about this long term! I'm sure any comments and observations will be very useful when it comes to internal refactor PRs.

---

I addressed the suggestion and added more context to the other comment.

parsely/src/main/java/com/parsely/parselyandroid/EngagementManager.kt
oguzkocer: Thanks a lot for the context!

> There's also a possibility, that the original implementation was enqueuing this event unintended - do you think it's what happened here?

I thought that might be possible - but the previous implementation looks pretty intentional.

---

This logic feels a little fragile and out of place, but as we discussed previously, it makes sense for this PR to closely match the previous logic. So, I don't think we need to do anything about that at the moment.

Adding a comment about what's happening _could_ be useful, but I'll admit that I still don't full get it even with your explanation. I'd need to dive deeper into what each component is doing and how it's utilized, but that's not really worth it at the moment.

---

@wzieba I don't have any other comments that needs addressing, so please feel free to proceed with merging the PR.

PR #96, opened 2023-11-29
Migrate `ParselyMetadata` and `ParselyVideoMetadata` to Kotlin. Improve API.

wzieba:


parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.java
```diff
@@ -18,7 +17,7 @@
 public class ParselyMetadata {
     ArrayList<String> authors, tags;
     String link, section, thumbUrl, title;
-    Calendar pubDate;
+    long publicationDateMilliseconds;
```
wzieba: I'd appreciate feedback on this change. I don't like introducing a breaking change to the library, but at the same time, I would rather not force users to use `java.util.Calendar` API - there's no need to attach to this specific class to share a moment in time.

I've decided to propose a simple, API-agnostic, straightforward and universal `long` property. The consequence of this is that user can now easily provide an incorrect value (e.g. timestamp in seconds), but I sense that it's not that likely (name of the property) and simpler API has more value. WDYT?

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.kt
```diff
@@ -21,13 +21,13 @@ open class ParselyMetadata
...
-    private val thumbUrl: String?,
-    private val title: String?,
-    private val publicationDateMilliseconds: Long?
+    private val authors: List<String>? = null,
+    @JvmField internal val link: String? = null,
+    private val section: String? = null,
+    private val tags: List<String>? = null,
+    private val thumbUrl: String? = null,
+    private val title: String? = null,
+    private val publicationDateMilliseconds: Long? = null
```
wzieba, line 30: Instead of this, I was exploring introducing a builder pattern, so Java clients won't have to provide `null` in place of nullable fields.

I decided not doing it, as it'd deviate more from the original implementation, making addressing the breaking change more difficult. Furthermore, it'd be much different than iOS API. So I think, the default values for arguments will be just fine - making the API nicer for Kotlin users and not changing it for Java users.

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.java
oguzkocer: Could you elaborate on why using `Calendar` would be undesirable - besides the fact that it may not be the particular API the clients are using?

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.kt
oguzkocer: As far as I know, using Builder pattern would mean that we're adding to the API not changing it. We'd still have the `ParselyMetadata` without any breaking changes - but there would also be a `ParselyMetadataBuilder` which returns `ParselyMetadata` when we call `.build()` on it.

That's all to say that, if Builder pattern is something you'd like to utilize, I don't think it conflicts with the current implementation.

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.kt
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Represents post metadata to be passed to Parsely tracking.
+ *
+ *
+ * This class is used to attach a metadata block to a Parse.ly pageview
+ * request. Pageview metadata is only required for URLs not accessible over the
+ * internet (i.e. app-only content) or if the customer is using an "in-pixel" integration.
+ * Otherwise, metadata will be gathered by Parse.ly's crawling infrastructure.
+ */
+open class ParselyMetadata
```
oguzkocer, line 12: I always get a little worried when I see `open` keyword, but I guess this is keeping the current design - so it's all good 😅

oguzkocer (commented):
@wzieba I left a couple replies to your line comments - but I don't have much else to add as this is a very straightforward PR. I'll mark this as approved once the conversations are resolved.

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.java
wzieba: The concern about different APIs used by the clients is the main one, but `Calendar` also feels heavier than what we need from this field, which is passing a timestamp.

But, as the majority of the usage is probably passing `Calendar.getInstance()` here, maybe it's not a big deal and backward compatibility and safer API design is better? I'm not sure now.

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.kt
wzieba: That's true - thanks. I've tunneled vision on constructor OR builder, but there's no reason to not introduce builder separately from the constructor. Still, for now, I'd leave this as it is and keep the builder for possible enhancement in the future.

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.kt
wzieba: True 😃 but yes, that's intentional.

parsely/src/main/java/com/parsely/parselyandroid/ParselyMetadata.java
oguzkocer: In my opinion, in most cases using _a_ type is better than not having a type at all. One obvious exception to that is when using a specific type requires the clients to add a new dependency. However, I believe `java.util.Calendar` is available without any external dependencies, so I don't see any issues about that.

Even if the type we choose is not exactly the type the clients would like to use, they can always convert to it on a need to basis as long as it is properly documented. In this case, `Calendar` is a well known and properly documented class, so I think it's a very good candidate to be used in external facing APIs.

wzieba: You're right - `Calendar` is a well-known and available dependency for every client. I agree it's easy to convert from any other API. Thank you for sharing your thoughts! - I brought back `Calendar` in 67a755b

oguzkocer: Thank you! 🙇‍♂️

oguzkocer (approved):

PR #99, opened 2023-12-13
Migrate `ParselyTracker` from Java to Kotlin

wzieba:
Best review after #100 

### Description

This PR migrates `ParselyTracker`, from Java to Kotlin. This class is the starting point for any user of this SDK.

Aside from the migration, this PR introduces also default arguments for tracking methods, to make calling this SDK from Kotlin more convenient, some code simplifications. This PR also introduces enforcing visibility modifiers for the whole library by setting `explicitApi` to `STRICT`.

### Testing
This PR does not require testing. Functional tests should assert the validity of basic flows.


parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,442 @@
...
+import java.util.UUID
+
+/**
+ * Tracks Parse.ly app views in Android apps
+ *
+ *
+ * Accessed as a singleton. Maintains a queue of pageview events in memory and periodically
+ * flushes the queue to the Parse.ly pixel proxy server.
+ */
+public open class ParselyTracker protected constructor(siteId: String, flushInterval: Int, c: Context) {
```
wzieba, line 33: This shouldn't be `public` and `open` but I haven't found any other way for now. If I introduced some interface, let's say `Tracker` to use it in `EngagementManagerTest` to create a fake implementation, I'd have to expose `enqueueEvent` to `public` - apparently, interface methods cannot be `internal`.

But I believe this shouldn't be a blocker - `ParselyTracker`, as previously written in Java, was "`open`" by default.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,424 @@
...
```
oguzkocer, line 33: Shouldn't we use an `internal` constructor since the class documentation states that it should be accessed through the singleton?

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,424 @@
...
+            return
+        }
+        val pageViewUuid = lastPageviewUuid
+        if (pageViewUuid == null) {
+            log("engagement session cannot start without calling trackPageview first")
+            return
+        }
+
+        // Cancel anything running
+        stopEngagement()
```
oguzkocer, line 207: I realize this is following the existing implementation. However, what do you think about at least logging that we are stopping the existing engagement?

If I was designing this from scratch, I'd probably use states to dictate what can and can't be called. So, instead of `ParselyTracker.sharedInstance().startEngagement()`, it would be something like `ParselyTracker.sharedInstance().engagement.start()` and `ParselyTracker.sharedInstance().engagement.stop()` where the `engagement` would be an enum that has started and stopped states. (at the minimum)

This type of design would not allow for an engagement to be started before a previous one is stopped. It also makes `private var engagementManager` property to be non-nullable.

Note that I only spent 5-10 minutes thinking about this, and didn't try it out - so I wouldn't be surprised if there are holes in it 😅

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,424 @@
...
+    /**
+     * Set a debug flag which will prevent data from being sent to Parse.ly
+     *
+     *
+     * Use this flag when developing to prevent the SDK from actually sending requests
+     * to Parse.ly servers. The value it would otherwise send is logged to the console.
+     *
+     * @param debug Value to use for debug flag.
+     */
+    public fun setDebug(debug: Boolean) {
```
oguzkocer, line 136: This, similar to my previous comment, follows the existing implementation, and I think it's a safe one for how it's used at the moment. However, it seems very risky to me to allow changing this flag after the tracker is initialized. In another implementation, it leaves room for the debug flag to corrupt the state - especially in the case of `debug` flag being initially turned on and then being turned off. Even in the current implementation, I think it's possible for this scenario to be interpreted as "dropped events".

In any case, this is not a big deal. However, whenever I see a `setSomething` function in a non-data class, I get a little worried. In my opinion, the more we avoid these setters, the safer the code will be 🤷‍♂️

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,424 @@
...
+        urlMetadata: ParselyMetadata? = null,
+        extraData: Map<String, Any>? = null,
+    ) {
+        if (url.isBlank()) {
+            log("url cannot be empty")
+            return
+        }
+
+        val pageViewUuid = generatePixelId()
+        lastPageviewUuid = pageViewUuid
```
oguzkocer, line 165: Nitpick:

```suggestion
        lastPageviewUuid = generatePixelId()
```

I think we can drop the extra value.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,424 @@
...
+     * @param urlRef      Referrer URL associated with this video view.
+     * @param urlMetadata Optional metadata for the URL -- not used in most cases. Only needed
+     * when `url` isn't accessible over the Internet (i.e. app-only
+     * content). Do not use this for **content also hosted on** URLs Parse.ly
+     * would normally crawl.
+     * @param extraData   A Map of additional information to send with the event.
+     */
+    public fun trackPageview(
+        url: String,
+        urlRef: String? = null,
```
oguzkocer, line 155:
```suggestion
        urlRef: String = "",
```

I suggest that we update all the `urlRef` parameters to use this ^ instead of the nullable value and remove the `orEmpty()`.

This may seem like a nitpick, but besides the simplification, it exposes what we are doing with this argument in case it's omitted, so there is some value in it. I guess it's fairly obvious what we'd be doing with a `null` value in this case, but as a general principle, I think it's a good idea to be transparent when we can.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: >  If I introduced some interface, let's say `Tracker` to use it in `EngagementManagerTest` to create a fake implementation, I'd have to expose `enqueueEvent` to `public` - apparently, interface methods cannot be `internal`.

Assuming that this is what we wanted to do, wouldn't it be more appropriate to use multiple interfaces in this case? We'd have a `Tracker` interface that we expose to `public` and another `internal` interface that has `enqueueEvent` function. `ParselyTracker` then will implement both and everything should work? Let me know if I am missing anything in this scenario.

oguzkocer (commented):
@wzieba I finished my first pass of this, but I might have more to add in my second pass as I want to think about it a bit. Hope you find the comments useful!

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: Good point, I agree and addressed in ea37072. Initially, I was super cautious about breaking public API, but on the other hand, now is the best moment to do so (we'll be releasing a new major version anyway).

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: True, I also didn't like the `setDebug` method. You make good points and I think it's a good time to get rid of it.

I've addressed it in 27fab68 and then I propose changing `debug` to `dryRun` - I think the later describes the intention much better. WDYT? 9d75f60

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: We could, but the Kotlin compiler can't infer that `lastPageviewUuid` is not null. I think this additional variable fits better than `!!` - it always feels like a code smell, and was designed to feel like one. WDYT?

image

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: Yes, that's a good point. I've addressed it in another PR https://github.com/Parsely/parsely-android/pull/101/files#diff-1e65066b482c03c14a63d8653fe95e6f2239ced194564de01d5a1b2d97a9e12eR33

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: Even with multiple interfaces, we can't have a method in interface that is `internal`. Even if the interface itself is `internal`

image

If we didn't add `internal`, then `ParselyTracker#enqueueEvent` would be exposed to API consumers, which is something we should omit.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: I see the benefits of the proposed design, but when trying to prepare a POC, I realized that it's getting more complex. The `startEngagment` has many dependencies as it creates a new `EngagementManager`.

For now, I've added the log: a8bf91d91b7a9da4a1ccfda6b83488410a65a6db

wzieba:
Thank you @oguzkocer for the review! Ready for the second round.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: Sounds good! I think we actually had a discussion about using `dryRun` in one of the previous PRs 😅

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: That's a good point. I completely missed the mutability and this was a pretty bad suggestion. Sorry about that!

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: I think there are still a few more `urlRef`s that use the nullable approach.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: It's very likely that I am missing something, but I was thinking something like this would work? 74a9f1857b3a0609e39df92e1cb08ee31e1028ba

Note that this ^ is not production quality code, I just put it together to make it easier to discuss. Let me know what you think!

---

Also, if we did go with such a design, I think we can make `ParselyTracker` internal as well and expose the shared instance some other way.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: Thank you, missed it completely 🤦 . Fixed in 697c051

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: Thank you! I really like your idea, haven't seen in this way myself. I also appreciate it clarifies the public contract better.

I've cherry-picked your comment and added some changes. Most notably, I've renamed `Tracker` interface to `ParselyTracker` to minimize breaking changes for the API consumers.

wzieba:
Thank you @oguzkocer for the suggestions! Applied them, ready for another round.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: I am glad you found it useful! 🥳

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        /**
+         * Singleton instance accessor. Note: This must be called after [.sharedInstance]
+         *
+         * @return The singleton instance
+         */
+        @JvmStatic
+        public fun sharedInstance(): ParselyTracker? {
+            return instance
+        }
```
oguzkocer, line 73: I am not sure if the additional `sharedInstance` method is helpful or not. I think the following is the intended usage:

* First call `sharedInstance({configuration})`
* Subsequently call `sharedInstance()`

Although this works - the better approach in my opinion is to:

* First call `sharedInstance({configuration})` and assign it to a property (preferably somewhere where you can inject it)
* Then only access the tracker through your local (to the client binary) property and not use the `sharedInstance()`

This approach is better in my opinion, because it doesn't deal with the nullability and somewhat forces the instantiation of the property to be done in a more proper place. It's also how a client would handle the tracker if it wasn't using a singleton pattern.

So, I wonder if we should not give them the option to use the first approach at all. What do you think?

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -0,0 +1,99 @@
...
+         * @return The singleton instance
+         */
+        @JvmStatic
+        @JvmOverloads
+        public fun sharedInstance(
+            siteId: String,
+            flushInterval: Int = DEFAULT_FLUSH_INTERVAL_SECS,
+            context: Context,
+            dryRun: Boolean = false,
+        ): ParselyTracker {
```
oguzkocer, line 91: A singleton that requires a configuration presents an interesting problem. What happens if the client tries to pass a different configuration after the instance is already initialized?

Currently, the secondary configuration is silently ignored. And, I could see a case where this can result in developers trying to figure out why their configuration change is not applied. Assuming that we don't want to allow updating the configuration after the initialization, I think we can give them a heads up about this in the documentation and by printing a log.

How do you feel about:
1. Adding a note to the documentation of this method, saying that this method should only be called once and subsequent calls will ignore its arguments and return the existing instance
2. Either add a generic warning about this ^ as a log every time we return the existing instance. Or - add a specific warning if this method is called with a different configuration as compared to the existing instance.

Note that an alternative approach that would avoid this whole issue would be to use a build time configuration instead of function parameters, but I don't think that's appropriate for this use case. Even if it was, that approach has its own issues - but I wanted to mention it for completeness-sake.

oguzkocer (approved):
@wzieba Thank you for addressing my previous feedback. I am glad that it was useful.

I had a one last look and wanted to bring up a couple points about the singleton. I am approving the changes just so this PR doesn't block you anymore, but I am curious to hear your thoughts on those.

In general, it's looking really good and the SDK is looking **significantly** better 🙇‍♂️

wzieba:
Thank you @oguzkocer for the comments about `ParselyTracker#sharedInstance`! I think it'll be the best to continue this discussion in a separate PR - merging this one, prepare changes in another PR and ping you there 🙂.

PR #101, opened 2023-12-14
Introduce binary compatibility validator

wzieba:
## Description

This PR introduces a binary compatibility validator to help reduce risks of introducing unwanted public API change.

### How to test

Please add or remove a public method from any class and then run `apiCheck` task. It should fail.


wzieba:
This PR introduces a development tool. It doesn't affect SDK consumers. I'll merge this without review as it's not dangerous in any way.

PR #105, opened 2024-01-18
Improve how `ParselyTracker` is accessed

wzieba:
## Description

This PR is addressing 2 review comments from a different PR: https://github.com/Parsely/parsely-android/pull/99#discussion_r1453739601 https://github.com/Parsely/parsely-android/pull/99#discussion_r1453748066

I identify 2 improvement opportunities here:
- confusing SDK initialization and re-initialization problem
- nullability of `sharedInstance()` getter and confusing usage

This PR should address both matters. I'll put details in comments.

## How to test
No need for manual tests.

### DIFF
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
@@ -110,7 +110,9 @@ public object ParselyTracker {
         context: Context,
         dryRun: Boolean = false,
     ) {
-        Logging.log("Parse.ly has been already initialized. Previous configuration will be overwritten.")
+        if (instance != null) {
+            Logging.log("Parse.ly has been already initialized. Previous configuration will be overwritten.")
+        }
         instance = ParselyTrackerInternal(siteId, flushInterval, context, dryRun)
     }
 }
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
@@ -83,17 +83,14 @@ public interface ParselyTracker {
          */
         @JvmStatic
         @JvmOverloads
-        public fun sharedInstance(
+        public fun init(
             siteId: String,
             flushInterval: Int = DEFAULT_FLUSH_INTERVAL_SECS,
             context: Context,
             dryRun: Boolean = false,
-        ): ParselyTracker {
-            return instance ?: run {
-                val newInstance = ParselyTrackerInternal(siteId, flushInterval, context, dryRun)
-                instance = newInstance
-                return newInstance
-            }
+        ) {
+            Logging.log("Parse.ly has been already initialized. Previous configuration will be overwritten.")
+            instance = ParselyTrackerInternal(siteId, flushInterval, context, dryRun)
         }
     }
 }
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
@@ -16,6 +16,7 @@
 package com.parsely.parselyandroid
 
 import android.content.Context
+import kotlin.jvm.Throws
 
 /**
  * Tracks Parse.ly app views in Android apps
@@ -25,52 +26,72 @@ import android.content.Context
  * flushes the queue to the Parse.ly pixel proxy server.
  */
 public interface ParselyTracker {
-    public val engagementInterval: Double?
-    public val videoEngagementInterval: Double?
-    public val flushInterval: Long
-
-    public fun engagementIsActive(): Boolean
-    public fun videoIsActive(): Boolean
-
-    public fun trackPageview(
-        url: String,
-        urlRef: String = "",
-        urlMetadata: ParselyMetadata? = null,
-        extraData: Map<String, Any>? = null,
-    )
-
-    public fun startEngagement(
-        url: String,
-        urlRef: String = "",
-        extraData: Map<String, Any>? = null
-    )
-
-    public fun stopEngagement()
-
-    public fun trackPlay(
-        url: String,
-        urlRef: String = "",
-        videoMetadata: ParselyVideoMetadata,
-        extraData: Map<String, Any>? = null,
-    )
-    public fun trackPause()
-    public fun resetVideo()
-    public fun flushEventQueue()
-    public fun flushTimerIsActive(): Boolean
 
     public companion object {
         private const val DEFAULT_FLUSH_INTERVAL_SECS = 60
         private var instance: ParselyTrackerInternal? = null
 
-        /**
-         * Singleton instance accessor. Note: This must be called after [.sharedInstance]
-         *
-         * @return The singleton instance
-         */
-        @JvmStatic
-        public fun sharedInstance(): ParselyTracker? {
-            return instance
+        @Throws(ParselyNotInitializedException::class)
+        private fun ensureInitialized(): ParselyTrackerInternal {
+            return instance ?: run {
+                throw ParselyNotInitializedException("Parse.ly client has not been initialized. Call ParselyTracker#init before using the SDK.");
+            }
         }
+
+        public val engagementInterval: Double?
+            @JvmStatic
+            get() = ensureInitialized().engagementInterval
+
+        public val videoEngagementInterval: Double?
+            @JvmStatic
+            get() = ensureInitialized().videoEngagementInterval
+        public val flushInterval: Long
+            @JvmStatic
+            get() = ensureInitialized().flushInterval
+
+        @JvmStatic
+        public fun engagementIsActive(): Boolean = ensureInitialized().engagementIsActive()
+
+        @JvmStatic
+        public fun videoIsActive(): Boolean = ensureInitialized().videoIsActive()
+
+        @JvmStatic
+        public fun trackPageview(
+            url: String,
+            urlRef: String = "",
+            urlMetadata: ParselyMetadata? = null,
+            extraData: Map<String, Any>? = null,
+        ): Unit = ensureInitialized().trackPageview(url, urlRef, urlMetadata, extraData)
+
+        @JvmStatic
+        public fun startEngagement(
+            url: String,
+            urlRef: String = "",
+            extraData: Map<String, Any>? = null
+        ): Unit = ensureInitialized().startEngagement(url, urlRef, extraData)
+
+        @JvmStatic
+        public fun stopEngagement(): Unit = ensureInitialized().stopEngagement()
+
+        @JvmStatic
+        public fun trackPlay(
+            url: String,
+            urlRef: String = "",
+            videoMetadata: ParselyVideoMetadata,
+            extraData: Map<String, Any>? = null,
+        ): Unit = ensureInitialized().trackPlay(url, urlRef, videoMetadata, extraData)
+
+        @JvmStatic
+        public fun trackPause(): Unit = ensureInitialized().trackPause()
+
+        @JvmStatic
+        public fun resetVideo(): Unit = ensureInitialized().resetVideo()
+
+        @JvmStatic
+        public fun flushEventQueue(): Unit = ensureInitialized().flushEventQueue()
+
+        @JvmStatic
+        public fun flushTimerIsActive(): Boolean = ensureInitialized().flushTimerIsActive()
 
         /**
          * Singleton instance factory Note: this must be called before [.sharedInstance]
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
@@ -30,7 +30,6 @@ public object ParselyTracker {
     private const val DEFAULT_FLUSH_INTERVAL_SECS = 60
     private var instance: ParselyTrackerInternal? = null
 
-    @Throws(ParselyNotInitializedException::class)
     private fun ensureInitialized(): ParselyTrackerInternal {
         return instance ?: run {
             throw ParselyNotInitializedException("Parse.ly client has not been initialized. Call ParselyTracker#init before using the SDK.");
--- a/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
+++ b/parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
@@ -25,93 +25,92 @@ import kotlin.jvm.Throws
  * Accessed as a singleton. Maintains a queue of pageview events in memory and periodically
  * flushes the queue to the Parse.ly pixel proxy server.
  */
-public interface ParselyTracker {
+public object ParselyTracker {
 
-    public companion object {
-        private const val DEFAULT_FLUSH_INTERVAL_SECS = 60
-        private var instance: ParselyTrackerInternal? = null
+    private const val DEFAULT_FLUSH_INTERVAL_SECS = 60
+    private var instance: ParselyTrackerInternal? = null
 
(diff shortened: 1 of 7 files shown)


parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -24,76 +25,94 @@ import android.content.Context
...
+    public fun init(
+        siteId: String,
+        flushInterval: Int = DEFAULT_FLUSH_INTERVAL_SECS,
+        context: Context,
+        dryRun: Boolean = false,
+    ) {
+        if (instance != null) {
+            Logging.log("Parse.ly has been already initialized. Previous configuration will be overwritten.")
         }
+        instance = ParselyTrackerInternal(siteId, flushInterval, context, dryRun)
```
wzieba, line 116: I took an inspiration for this method from Sentry SDK ([link](https://github.com/getsentry/sentry-java/blob/2465853df157c4ad881166804873989c47c10f4d/sentry/src/main/java/io/sentry/Sentry.java#L218-L230)).

I think `ParselyTracker.init` is a clear name for the initialization method and it won't cause any doubts.

Citing [your commen](https://github.com/Parsely/parsely-android/pull/99#discussion_r1453748066)t @oguzkocer 

> A singleton that requires a configuration presents an interesting problem. What happens if the client tries to pass a different configuration after the instance is already initialized? (...) Assuming that we don't want to allow updating the configuration after the initialization, I think we can give them a heads up about this in the documentation and by printing a log.

I'm not entirely sure if we want or do not want to allow users to reinitialize the tracker. Previously we didn't allow, but maybe there's no reason to?

My only concern currently is: what happens if a user started the engagement session and then re-initialized SDK? I guess there might be a memory leak here, but haven't tested it _yet_.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -24,76 +25,94 @@ import android.content.Context
  * Accessed as a singleton. Maintains a queue of pageview events in memory and periodically
  * flushes the queue to the Parse.ly pixel proxy server.
  */
-public interface ParselyTracker {
+public object ParselyTracker {
```
wzieba, line 28: I've decided to propose API change from usages like
```
ParselyTracker.sharedInstance().trackPageview()
```

to
```
ParselyTracker.trackPageview()
```

I think the second API is more expected (looking at API designs of e.g. Sentry or Crashlytics). To address possible `null` when calling SDK methods before `init` method, I've decided to propose "fail fast" approach of throwing our custom exception which clearly points the user to the problem.

wzieba:
hi @oguzkocer 👋 Here's my proposal to address comments from the last review. WDYT about them?

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: On the second thought, I've decided to disable re-initialization (35e9cdda5de5f63e1e5596bbcbf9d9b5c4626764). The SDK doesn't allow doing this now, and I see no good arguments to allow it.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: I think this approach should work well - but I am not sure why we need to give up on the interface. Shouldn't we keep the previous interface and implement it for this object to make mocking possible?

parsely/src/main/java/com/parsely/parselyandroid/ParselyTrackerInternal.kt
```diff
@@ -316,7 +316,7 @@ internal class ParselyTrackerInternal internal constructor(
      * Any usage of this method is safe to remove and will have no effect. Keeping for backwards compatibility.
      */
     @Deprecated("The SDK now automatically flushes the queue on app lifecycle events. Any usage of this method is safe to remove and will have no effect")
-    override fun flushEventQueue() {
+    fun flushEventQueue() {
```
oguzkocer, line 319: I think I might have missed this in previous PR reviews. Since we already have a lot of breaking changes, should we simply remove this no-op function as well? It's probably OK either way, but I figure we might as well start fresh after all these breaking changes.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -24,76 +25,95 @@ import android.content.Context
...
+    @JvmStatic
+    @JvmOverloads
+    public fun init(
+        siteId: String,
+        flushInterval: Int = DEFAULT_FLUSH_INTERVAL_SECS,
+        context: Context,
+        dryRun: Boolean = false,
+    ) {
+        if (instance != null) {
+            Logging.log("Parse.ly SDK has been already initialized. Reinitialization is not supported.")
```
oguzkocer, line 114: Now that this is in an `init` method, I think an exception is more appropriate than a log.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTrackerInternal.kt
wzieba: Yes, definitely. I was considering doing this in a separate PR, but on the other hand, I can do this here. 59f21aa29ca3b5c05aa576017b442db7dd6f05bb

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: > I am not sure why we need to give up on the interface

I realized that the interface won't be needed anymore. Looking from the API consumer perspective, I can't find a good justification for it. If it'll be needed in the future for internal usage, we can always revert it. But for now, I would not introduce it to keep things simpler. WDYT?

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: I've introduced it in 1ddfda1 and tested in 0478d9e.

I'm struggling with my thoughts though - I understand the reasoning of signaling to the API consumers that they're doing something that they shouldn't as fast as possible. On the other hand, I'm not entirely sure if the analytics SDK should crash the app in case of such "not destructive" event.

Nevertheless, I'm not convinced strongly to any side, so I'm aligning to your comment. Just wanted to share this thought.

oguzkocer: That's a good point - I treated this as if it was something internal. As a consumer, I like it crashing instead of silent failures, but I could see other devs preferring a non-destructive approach to this issue. So, I am not sure which is better either.

I'll leave the final decision to you.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: If the consumer wants to write a unit test, wouldn't an interface be necessary to mock the tracker? I could totally be missing something - as we discussed - I don't do much active Kotlin development and maybe it's possible to mock it as is.

wzieba: I'm not sure the interface will be possible with such a design of invoking methods statically. Methods declared in the interface would have to be either `@JvmStatic` or not. 

**With @JvmStatic**
If we had an interface like in this POC commit: f4a618e540aea0091274c3f2751224b7c3e4fdba then we can't call method as member: 
image

making this still untestable.

**Without @JvmStatic**
Alternatively, if we remove `@JvmStatic`, then we can call method as member function, but can't as static method.

image


I'd like to keep static calls for the convenience of the clients and reduce the overhead of migration, even if it makes this part of the code untestable.

---

The alternative for the clients would be to prepare a wrapper - just like we do in other projects e.g. [here](https://github.com/woocommerce/woocommerce-android/blob/trunk/libs/cardreader/src/main/java/com/woocommerce/android/cardreader/internal/wrappers/TerminalWrapper.kt).

WDYT? We can always revert to some variant of `sharedInstance` method, but I'm not sure if it's worth it. E.g. Sentry Java SDK, SDK with some simillarities to this one, doesn't provide an interface to its entrypoint ([link](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Sentry.java#L40)).

wzieba: Actually, I don't think POC from https://github.com/Parsely/parsely-android/commit/f4a618e540aea0091274c3f2751224b7c3e4fdba is possible at all: there's declaration clash between the interface and its companion object

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: I think we can leave the exception for now and act on it if we'll have feedback about this from SDK consumers. It'll be easier to update SDK from "fail fast" to "do not fail" rather than other way around.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -24,76 +25,98 @@ import android.content.Context
...
-    public fun engagementIsActive(): Boolean
-    public fun videoIsActive(): Boolean
+    @JvmStatic
+    public fun engagementIsActive(): Boolean = ensureInitialized().engagementIsActive()
 
+    @JvmStatic
+    public fun videoIsActive(): Boolean = ensureInitialized().videoIsActive()
+
+    @JvmStatic
+    @JvmOverloads
```
wzieba, line 58: Sorry for one more change in this PR @oguzkocer . I've added `@JvmOverloads` annotations so the API will be nicer to work for Java consumers
image

I thought I added it already before, but can't find a place. It'll be easier to track such changes after #101 is merged, which I plan to make ready for review after this PR.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
oguzkocer: Couldn't we have the old design with the interface and have the new static objects on top as a wrapper? The wrapper could simply call the `ParselyTracker.sharedInstance()` underneath? I didn't give this a try though, so hopefully I am not missing anything.

---

As nice as it's to be able to call `ParselyTracker.trackPageview` directly, I feel like the benefit here is purely cosmetic. On the other hand, using an interface design has the benefit of making it mockable and unit testable by consumers. So, for me, the interface design is the clear winner.

I _think_ we could do both, as suggested at the top of this comment - but if it was up to me, I'd only go with the interface design and not add the static object at all because having multiple APIs could be confusing.

Having said all that, if you want to go ahead with the static object design, I respect that decision and I can review the PR as is. Also, if you want a third opinion, feel free to ping other developers as well.

---

P.S: I think it's a great idea to look at other 3rd party SDKs, but I personally never liked Sentry's approach. I could totally be the odd man out, so please take that possibility into consideration.

wzieba: Gotcha, I've brought back the interface and `sharedInstance` method in fa7a23e8efe79a36d39f032c412e2a0e7344c850

I agree, the benefit would be only cosmetic and would also cause more rework on client side. I also agree that a mockable interface is a nice thing to have for consumers. I hardly remember an Android SDK providing such, I'm not sure why, though. Or maybe I just missed it.

>  I'd only go with the interface design and not add the static object at all because having multiple APIs could be confusing.

I'm afraid, lack of static object will cause significant refactoring effort for the consumers because of the need of passing `ParselyTracker` object. I can imagine this API used in places like e.g. Android Services where it's sometimes not trivial to pass an object directly. Also, I think usually logging SDK I've seen have static APIs: `Crashlytics`, `android.util.Log`, `Log4j`

So, to sum up, with the proposal introduced in: fa7a23e8efe79a36d39f032c412e2a0e7344c850 the SDK will have 2 ways of using it:

**Via interface**
This usage allows for unit testing `SomeClass` by mocking `ParselyTracker` interface

```kotlin

fun onCreate(...) {
    ParselyTracker.init(...)
    val tracker = ParselyTracker.sharedInstance()

    val someClass = SomeClass(tracker)
    someClass.openArticle()
}

class SomeClass(val tracker: ParselyTracker) {
    
    fun openArticle() {
        ...
        tracker.startEngagement()
        ...
    }
}
```

**Via static object**
This makes `SomeClass` more difficult to test, but the API easier to use and backward compatible

```kotlin
fun onCreate(...) {
    ParselyTracker.init(...)

    val someClass = SomeClass()
    someClass.openArticle()
}

class SomeClass() {
    
    fun openArticle() {
        ...
        ParselyTracker.sharedInstance().startEngagement()
        ...
    }
}
```

oguzkocer: I assume the `Via static object` would use `ParselyTracker.startEngagement()` instead of `ParselyTracker.sharedInstance().startEngagement()`?

Also, I think it'll be safe to mix and match the usages, could you confirm that?

---

It's not ideal to have 2 different ways to achieve the same thing, but I understand the arguments for it. If you believe this is the way to go, I am good with it.

oguzkocer:
@wzieba Is this PR a blocker for you? Can I review it again on Friday?

I am trying to finish something until EOD today, so if this is not a blocker for you, I can pay more attention to the review on Friday.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
wzieba: > I assume the Via static object would use ParselyTracker.startEngagement() instead of ParselyTracker.sharedInstance().startEngagement()?

No, `ParselyTracker.startEngagement()` will not work anymore (after the recent commit: https://github.com/Parsely/parsely-android/commit/fa7a23e8efe79a36d39f032c412e2a0e7344c850). To use SDK statically, one would need to use `ParselyTracker.sharedInstance().startEngagement()` now.

I see that the name of this approach is confusing, sorry. What I meant was "via statically exposed object" and this object is `ParselyTrackerInternal` (hidden behind `ParselyTracker` interface).

> Also, I think it'll be safe to mix and match the usages, could you confirm that?

It'll be safe to mix usages - the `ParselyTracker` will make sure that there's only one instance of the `ParselyTrackerInternal` ever created.

wzieba:
@oguzkocer no worries, it's completely fine. Thanks for the heads up!

parsely/src/main/java/com/parsely/parselyandroid/ParselyTrackerInternal.kt
```diff
@@ -6,7 +6,7 @@ import androidx.lifecycle.LifecycleEventObserver
 import androidx.lifecycle.ProcessLifecycleOwner
 import java.util.UUID
 
-internal class ParselyTrackerInternal internal constructor(
+public class ParselyTrackerInternal internal constructor(
```
oguzkocer, line 9: Any chance this is an accidental change? 😓 If not, do you mind explaining why this needs to be public?

oguzkocer (commented):
@wzieba I left one comment, but other than that I think this is OK to merge as is. I am still tiny bit concerned about having multiple ways to use the SDK, but since you confirmed that it's safe to mix and match between them, I can't think of any major practical issues with it.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTrackerInternal.kt
wzieba: Yes, definitely accidental change.. thank you for noticing this 🙇

PR #109, opened 2024-01-30
Remove unneeded methods from api

wzieba:
## Description

This PR removes some methods from the public API. Those methods are not needed to use the SDK and were making the API less readable. Internal communication: p1706537769574079-slack-C0533SEJ82H

### Reflection invocation

Those methods, although not needed for public API, are useful in the scope of the demo app. That's why I've decided to keep them in `ParselyTrackerInternal` and invoke them using reflection.

### Testing

Install the `example` app and tap on each button. The app should not crash.

### DIFF
--- a/example/src/main/java/com/example/MainActivity.java
+++ b/example/src/main/java/com/example/MainActivity.java
@@ -50,7 +50,7 @@ public class MainActivity extends Activity {
                 TextView[] v = (TextView[]) msg.obj;
                 TextView iView = v[0];
                 if (ParselyTracker.sharedInstance().flushTimerIsActive()) {
-                    iView.setText(String.format("Flush Interval: %d", ParselyTracker.sharedInstance().getFlushInterval()));
+                    iView.setText(String.format("Flush Interval: %d", getFlushInterval()));
                 } else {
                     iView.setText("Flush timer inactive");
                 }
@@ -150,6 +150,10 @@ public class MainActivity extends Activity {
     @Nullable
     private Double getVideoEngagementInterval() {
         return (Double) invokePrivateMethod("getVideoEngagementInterval");
+    }
+
+    private long getFlushInterval() {
+        return (long) invokePrivateMethod("getFlushInterval");
     }
 
     private Object invokePrivateMethod(String methodName, Object... args) {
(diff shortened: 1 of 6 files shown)


parsely/api/parsely.api
```diff
@@ -14,11 +14,6 @@ public final class com/parsely/parselyandroid/ParselyNotInitializedException : j
```
wzieba, line 14: This file comes from #101 and highlights how the public API changed.

example/src/main/java/com/example/MainActivity.java
```diff
@@ -117,13 +127,51 @@ public void trackPlay(View view) {
...
     public void trackPause(View view) {
         ParselyTracker.sharedInstance().trackPause();
     }
 
-    public void trackReset(View view) {ParselyTracker.sharedInstance().resetVideo(); }
+    public void trackReset(View view) {
+        ParselyTracker.sharedInstance().resetVideo();
+    }
+
+    private boolean engagementIsActive() {
```
iangmaia, line 142: Are these private API calls really useful in the demo app? Meaning: does it make sense to show this info given it's not going to be available for clients?
I get that the intention might be just showing some debug info, but if so, perhaps good to at least add a comment highlighting that these are private methods and not intended to be used by users in general? 🤔

wzieba: > Are these private API calls really useful in the demo app?

I think yes - it might be helpful to see how the internal parameters of the SDK behave after certain actions to see how it works. Especially, as things happen here after intervals 10+ seconds.

But I see that it's the "example" app, so it should be giving a good example of the integration. I agree that in the current form it could be misleading.

In 003d6fb760f692812ca21f612324610cb24bf5e0 I've extracted all calls to a separate class and added comments. WDYT?

iangmaia: > But I see that it's the "example" app, so it should be giving a good example of the integration
🎯 Exactly!

> In https://github.com/Parsely/parsely-android/commit/003d6fb760f692812ca21f612324610cb24bf5e0 I've extracted all calls to a separate class and added comments. WDYT?

This is a good compromise IMO, and it makes it clear to clients that this is just for debugging 👍 Thanks for taking the time to address it!

iangmaia (approved):

Issue #115, opened 2024-04-02
[feature request] Support multiple Site IDs within one app


thompsonjoshua:
A customer has one app with content from multiple subdomains that each have their own dashboard. They are looking for a way to load the correct Parse.ly Site ID to the corresponding url.

PR #116, opened 2024-04-05
Support `siteId` per tracking request

wzieba:
Closes: #115 

### Description

This PR introduces a possibility for SDK users to provide a custom `siteId` whenever they record a pixel/track event.

This will also align to iOS, which supports this feature already

https://github.com/Parsely/AnalyticsSDK-iOS/blob/9b9db5478996582fdbd07d5aa4a2da0134f9ffe7/Sources/ParselyTracker.swift#L47-L48

#### Design

I've decided to propose more verbose API and introduce `SiteIdSource` instead of a `String` which is empty by default. I think this way it's easier for SDK users to understand the behavior.

#### Backwards compatibility

Because we cannot introduce `@JvmOverloads` on interface methods, this change will be breaking for Java users. But I think this is acceptable - I plan to release `4.1.0` after this (not bumping major version). In the case of Java users, addressing breaking changes will be as easy as providing `SiteIdSource.Default.INSTANCE` in tracking methods.

### Test
This PR introduces unit tests and functional test case for engagement manager case. We can also test this with example app:

1. Run the example app
2. Provide some value in "Custom site id" edit text
3. Tap on "Track URL", then press "Start engagement"
4. Observe in logs that after ~15 seconds you should see logs that would be sent to the backend. Make sure they have `idsite` equal to the custom id you provided in step 2.

### DIFF
--- a/example/src/main/java/com/example/MainActivity.java
+++ b/example/src/main/java/com/example/MainActivity.java
@@ -11,7 +11,7 @@ import android.widget.TextView;
 
 import com.parsely.parselyandroid.ParselyTracker;
 import com.parsely.parselyandroid.SiteIdSource;
-import com.parsely.parselyandroid.internal.ParselyTrackerInternal;
+import com.parsely.parselyandroid.ParselyTrackerInternal;
 import com.parsely.parselyandroid.ParselyVideoMetadata;
 
 import java.util.ArrayList;
--- a/example/src/main/java/com/example/MainActivity.java
+++ b/example/src/main/java/com/example/MainActivity.java
@@ -8,6 +8,7 @@ import android.view.View;
 import android.widget.TextView;
 
 import com.parsely.parselyandroid.ParselyTracker;
+import com.parsely.parselyandroid.SiteIdSource;
 import com.parsely.parselyandroid.internal.ParselyTrackerInternal;
 import com.parsely.parselyandroid.ParselyVideoMetadata;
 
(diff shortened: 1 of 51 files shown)


wzieba:
hi @iangmaia 👋 would you be ok reviewing this PR?

parsely/src/main/java/com/parsely/parselyandroid/EventsBuilder.kt
```diff
@@ -24,15 +22,19 @@ internal class EventsBuilder(
         action: String,
         metadata: ParselyMetadata?,
         extraData: Map<String, Any>?,
-        uuid: String
+        uuid: String,
+        siteIdSource: SiteIdSource,
```
iangmaia, line 26: Parameter docs for these two are missing.
Perhaps adding more details specially on how `siteIdSource` works with its default can also be helpful.

parsely/src/main/java/com/parsely/parselyandroid/ParselyTracker.kt
```diff
@@ -36,12 +36,14 @@ public interface ParselyTracker {
      *                    content). Do not use this for **content also hosted on** URLs Parse.ly
      *                    would normally crawl.
      * @param extraData   A Map of additional information to send with the event.
+     * @param siteIdSource The source of the site ID to use for the event.
```
iangmaia, line 39: Perhaps it is worth it to add / repeat / emphasize a bit more info on siteId and what's the default behavior?

wzieba:
Thanks @iangmaia ! I've added and extended comments, ready for another round 👍

example/src/main/java/com/example/MainActivity.java
```diff
@@ -31,7 +35,7 @@ protected void onCreate(Bundle savedInstanceState) {
         setContentView(R.layout.activity_main);
 
         // initialize the Parsely tracker with your site id and the current Context
-        ParselyTracker.init("example.com", 30, this, true);
+        ParselyTracker.init(DEFAULT_SITE_ID, 30, this, true);
```
iangmaia, line 38: I wonder if `ParselyTracker.init()` shouldn't also receive a `SiteIdSource` 🤔 
Are there reasons not to?

wzieba: Because we need *some* site id to be defined. If `ParselyTracker.init()` received `SiteIdSource`, then API consumer could pass `SiteIdSource.Default`, which would have to be supported and it's not a valid state.

Alternatively, we could make `ParselyTracker.init()` receive `SiteIdSource.Custom`, but I think it only shadows implementation and causes confusion as it can raise a question: "custom" to what? In `ParselyTracker.trackPageview` it's "custom" to the value provided in initialization, so it makes sense there.

iangmaia: > Because we need _some_ site id to be defined. If `ParselyTracker.init()` received `SiteIdSource`, then API consumer could pass `SiteIdSource.Default`, which would have to be supported and it's not a valid state.
> 
> Alternatively, we could make `ParselyTracker.init()` receive `SiteIdSource.Custom`, but I think it only shadows implementation and causes confusion as it can raise a question: "custom" to what? In `ParselyTracker.trackPageview` it's "custom" to the value provided in initialization, so it makes sense there.

Got it! Makes sense 👍

iangmaia (approved):

Issue #119, opened 2024-12-06
Fatal Exception: java.lang.OutOfMemoryError


rostyslavfilipchuk:
Hello. Crashlytics reports a low amount of such crashes

```
Crash com.parsely.parselyandroid  Logging.java:15 
Logging.log 
java.lang.OutOfMemoryError
```
Reports from
Android 14, Samsung  Galaxy S21 Ultra 5G

Parsely version : 4.1.0

```
Fatal Exception: java.lang.OutOfMemoryError
Failed to allocate a 68817440 byte allocation with 8388608 free bytes and 58MB until OOM, target footprint 215655704, growth limit 268435456
java.lang.StringFactory.newStringFromUtf16Bytes (StringFactory.java)
java.lang.StringUTF16.newString (StringUTF16.java:1141)
java.lang.StringBuilder.toString (StringBuilder.java:475)
java.util.Formatter.toString (Formatter.java:2540)
com.parsely.parselyandroid.Logging.log (Logging.java:15)
com.parsely.parselyandroid.FlushQueue$invoke$1.invokeSuspend (FlushQueue.kt:41)
kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:108)
kotlinx.coroutines.internal.LimitedDispatcher$Worker.run (LimitedDispatcher.java:115)
kotlinx.coroutines.scheduling.TaskImpl.run (Tasks.kt:103)
kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.java:584)
kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask (CoroutineScheduler.kt:793)
kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker (CoroutineScheduler.kt:697)
kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt:684)
```

Anything I can do to stop it?

wzieba:
Thank you for the report @rostyslavfilipchuk !  #120 should address this problem, you can expect the fix to be shipped in `4.1.1`.

Issue #83, opened 2023-10-26
Use Android logger instead of system's output stream


wzieba:
Originally: https://github.com/Parsely/parsely-android/pull/82#discussion_r1372865532

Allows for more fine-grained level of logging.

PR #120, opened 2025-01-10
Use `android.util.Log` for logging

wzieba:
### Description

This way, the SDK has a control over which logs will be used in production or debug variants of SDK users. As on outcome, this change reduces number of logs that the SDK is producing in production apps.

Closes: #83
Closes: #119 

The #119 issue is a result of `StringBuilder` trying to allocate a significant part of available memory. This is probably a result of a case, in which consumer app tries to report many events: either due to device being offline for a long period or a misconfiguration of the SDK. This PR will stop logging whole payloads in production apps, what will likely address the issue.

### Testing

Not necessary: both unit and functional tests pass. If still needed, one can open an example app and validate the logs via Logcat.

image


parsely/src/main/java/com/parsely/parselyandroid/FlushQueue.kt
```diff
@@ -33,22 +32,21 @@ internal class FlushQueue(
...
-                log("POST Data %s", jsonPayload)
-                log("Requested %s", ParselyTrackerInternal.ROOT_URL)
+                Log.d("Sending request with ${eventsToSend.size} events")
+                Log.d("POST Data $jsonPayload")
+                Log.d("Requested ${ParselyTrackerInternal.ROOT_URL}")
                 restClient.send(jsonPayload)
                     .fold(
                         onSuccess = {
-                            log("Pixel request success")
+                            Log.i("Pixel request success")
```
ParaskP7, line 45: Question (❓): This is the only `Log.i("Xyz")` that I found. Just for my FYI, why did you chose INFO instead of DEBUG here? 🤔

parsely/src/main/java/com/parsely/parselyandroid/Log.kt
```diff
@@ -0,0 +1,34 @@
+package com.parsely.parselyandroid
+
+import android.util.Log as AndroidLog
+
+internal object AndroidLogWrapper : Log {
```
ParaskP7, line 5: Minor (🔍): Why the `Android` prefix, wouldn't `LogWrapper` be enough? 🤔

PS: Nothing serious, I just wonder. 😊

ParaskP7 (approved):
👋 @wzieba !

I have reviewed this PR as per the instructions (quickly tested it as well), everything works as expected, great job and much better logging indeed, awesome! 🌟

parsely/src/main/java/com/parsely/parselyandroid/FlushQueue.kt
wzieba: I wanted this log to be kept in the production variants of consumer apps (`DEBUG` level logs are stripped)

ParaskP7: 👍 if that was indeed intentional, I just wasn't sure about it! 😊

parsely/src/main/java/com/parsely/parselyandroid/Log.kt
wzieba: It's a wrapper around Android's `Log` so I went with `AndroidLogWrapper`.

ParaskP7: Just because we are anyway in the context of Android, to me this was a given, thus redundant, words! 😅

PR #124, opened 2026-05-15
Add trackConversion for sending conversion events

randyriback:
## Summary
- Adds `ParselyTracker.trackConversion(url, conversionType, conversionLabel, ...)` for firing conversion events (newsletter signups, subscriptions, purchases, link clicks, lead capture, custom). Internally constructs an event with `action="conversion"` and merges `_conversion_type` and `_conversion_label` into `extra_data`, which the Parse.ly conversions backend (`conversions-realtime`) extracts and uses as the goal identity.
- New `ConversionType` enum mirrors the categories accepted server-side (`newsletter_signup`, `lead_capture`, `link_click`, `subscription`, `purchase`, `custom`).
- Reuses the existing `EventsBuilder` / `InMemoryBuffer` / `FlushQueue` / `ParselyAPIConnection` pipeline — zero changes to the network layer, the buffer, or the flush manager.
- Example app gains two sandbox buttons used during end-to-end verification.

This is the Android counterpart to the iOS SDK's `trackConversion` (parallel PR on AnalyticsSDK-iOS).

### Unit tests (all passing)
- [x] `EventsBuilderTest.when building conversion event, then build the correct one` — verifies the built event has `action == "conversion"`, the right `url` and `idsite`, no `pvid` / `vsid`, and that `_conversion_type` / `_conversion_label` plus caller-supplied keys land in `data`.
- [x] `ParselyTrackerTest.given tracker initialized, when calling trackConversion, do not throw any exception` — smoke test for the public entry point (mirrors the existing call-doesn't-throw tests for other tracking methods).

Full suite (`./gradlew :parsely:testDebugUnitTest`): EventsBuilderTest 11 tests / ParselyTrackerTest 5 tests / 0 failures.

### Manual end-to-end (Android emulator: Pixel 5, android-34, default arm64-v8a)
- [x] Built and installed the example app on the emulator.
- [x] Flipped `dryRun` to `false` temporarily in `MainActivity.java` to exercise the real network path.
- [x] Tapped **Track Pageview (sandbox)** and **Track Conversion (sandbox)** against the `sandbox.joshhanson.io` apikey.
- [x] Verified the on-wire JSON via `adb logcat Parsely:V '*:S'`:
  ```json
  {
    "action": "conversion",
    "url": "https://sandbox.joshhanson.io/path/test-conversion2",
    "idsite": "sandbox.joshhanson.io",
    "data": {
      "_conversion_type": "subscription",
      "_conversion_label": "android_sdk_smoke_test",
      "plan": "weekly",
      "source": "android_demo_app",
      "os": "android",
      "os_version": "34",
      "parsely_site_uuid": "...",
      "ts": 1778782087996
    }
  }
### Testing Screenshots
<img width="782" height="1744" alt="Xnip2026-05-14_20-27-24" src="https://github.com/user-attachments/assets/65e59701-fbb1-44f5-8a46-79dfb6945151" />

<img width="973" height="434" alt="Screenshot 2026-05-14 at 8 46 39 PM" src="https://github.com/user-attachments/assets/2cddaae4-3065-4b36-8d05-cb559b59f347" />

### DIFF
--- a/example/src/main/java/com/example/MainActivity.java
+++ b/example/src/main/java/com/example/MainActivity.java
@@ -9,6 +9,7 @@ import android.view.View;
 import android.widget.EditText;
 import android.widget.TextView;
 
+import com.parsely.parselyandroid.ConversionType;
 import com.parsely.parselyandroid.ParselyTracker;
 import com.parsely.parselyandroid.SiteIdSource;
 import com.parsely.parselyandroid.ParselyTrackerInternal;
@@ -138,6 +139,34 @@ public class MainActivity extends Activity {
 
     public void trackReset(View view) {
         ParselyTracker.sharedInstance().resetVideo();
+    }
+
+    // --- Sandbox buttons for trackConversion smoke testing against sandbox.joshhanson.io.
+    // Both fire against the same URL so the conversions topology has a pageview to attribute to.
+    private static final String SANDBOX_SITE_ID = "sandbox.joshhanson.io";
+    private static final String SANDBOX_URL = "https://sandbox.joshhanson.io/path/test-conversion2";
+
+    public void trackSandboxPageview(View view) {
+        final Map<String, Object> extraData = new HashMap<>();
+        extraData.put("source", "android_demo_app");
+        ParselyTracker.sharedInstance().trackPageview(
+                SANDBOX_URL, "", null, extraData, new SiteIdSource.Custom(SANDBOX_SITE_ID)
+        );
+    }
+
+    public void trackSandboxConversion(View view) {
+        final Map<String, Object> extraData = new HashMap<>();
+        extraData.put("plan", "weekly");
+        extraData.put("source", "android_demo_app");
+        ParselyTracker.sharedInstance().trackConversion(
+                SANDBOX_URL,
+                ConversionType.SUBSCRIPTION,
+                "android_sdk_smoke_test",
+                "",
+                null,
+                extraData,
+                new SiteIdSource.Custom(SANDBOX_SITE_ID)
+        );
     }
 
     private SiteIdSource getSiteId() {
--- a/example/src/main/java/com/example/MainActivity.java
+++ b/example/src/main/java/com/example/MainActivity.java
@@ -101,7 +101,21 @@ public class MainActivity extends Activity {
         //       the post has an internet-accessible URL, we will crawl it. urlMetadata is only used
         //       in the case of app-only content that we can't crawl.
         ParselyTracker.sharedInstance().trackPageview(
-                "http://example.com/article1.html", "http://example.com/", null, null, getSiteId()
+                getUrl(), "http://example.com/", null, null, getSiteId()
+        );
+    }
+
+    public void trackConversion(View view) {
+        final Map<String, Object> extraData = new HashMap<>();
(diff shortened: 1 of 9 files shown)


example/src/main/java/com/example/MainActivity.java
```diff
@@ -140,6 +141,34 @@ public void trackReset(View view) {
         ParselyTracker.sharedInstance().resetVideo();
     }
 
+    // --- Sandbox buttons for trackConversion smoke testing against sandbox.joshhanson.io.
+    // Both fire against the same URL so the conversions topology has a pageview to attribute to.
+    private static final String SANDBOX_SITE_ID = "sandbox.joshhanson.io";
+    private static final String SANDBOX_URL = "https://sandbox.joshhanson.io/path/test-conversion2";
```
wzieba: The idea of `example` app is to sho clients how to use the SDK. I think the idea of hardcoded internal `sandbox` will be confusing for anynone outside the team.

The app already has an `EditText` for `site id`, so we could extend that pattern: add `EditText` for URL  (defaulting to example.com) at the top of the button list. The sandbox buttons could then be replaced with generic "Track Pageview" (already exists) and "Track Conversion" (new one) buttons that use those fields.

This way internal testing still works (just type in `sandbox.joshhanson.io`), but the default state of the app makes sense to clients.

parsely/src/main/java/com/parsely/parselyandroid/ConversionType.kt
```diff
@@ -0,0 +1,15 @@
...
+ * backend. Use [CUSTOM] for conversions that don't fit one of the named categories.
+ */
+public enum class ConversionType(internal val wireValue: String) {
+    NEWSLETTER_SIGNUP("newsletter_signup"),
+    LEAD_CAPTURE("lead_capture"),
+    LINK_CLICK("link_click"),
+    SUBSCRIPTION("subscription"),
+    PURCHASE("purchase"),
+    CUSTOM("custom"),
+}
```
wzieba, line 15: ❓ Just to make sure: this is a set of constants by design and we don't want to allow clients to declare their own `ConversionType`, right?

wzieba (commented):
Looks good! Just one suggestion regarding `example` app and one minor question.

parsely/src/main/java/com/parsely/parselyandroid/ConversionType.kt
randyriback: Exactly, @wzieba -- these are the expected conversion types (`"custom"` is supported but not documented): https://docs.parse.ly/installation-resources/add-on-features/set-up-conversion-tracking/#h-terminology

example/src/main/java/com/example/MainActivity.java
randyriback: Thanks @wzieba ! Refactored per your suggestion. The example app now has:

- A custom_url EditText next to the existing custom_site_id field (defaults to http://example.com/article1.html when empty).
- The existing "Track URL" button reads from custom_url.
- One new generic "Track Conversion" button that uses both fields, with ConversionType.SUBSCRIPTION + label "demo_conversion" hardcoded so the API surface stays visible to readers.

wzieba (approved):
Thanks @randyriback, LGTM 👍

